Issue #4, opened 2024-07-10
System related services


borditamas:
- System discovery service
- System mgmt service
- System DB service

PR #18, opened 2024-09-02
Kj/system

Katinka666:
related #4


serviceregistry/pom.xml
```diff
@@ -43,4 +43,16 @@
...
+			<groupId>eu.arrowhead</groupId>
+			<artifactId>arrowhead-common-utils</artifactId>
+			<version>5.0.0</version>
+		</dependency>
+		<dependency>
+			<groupId>eu.arrowhead</groupId>
+			<artifactId>arrowhead-data-transfer-objects</artifactId>
+			<version>5.0.0</version>
+		</dependency>
+	</dependencies>
```
rbocsi: This shouldn't be here. These dependencies are already in the parent's pom.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/ManagementAPI.java
```diff
@@ -178,15 +182,101 @@ public void removeDevices(final @RequestParam List<String> names) {
...
-	// query-systems POST /systems (query param verbose)
-	// * paging: page, size, direction, sort
-	// * filter to: name list, metadata requirement list, version list, address list, address type, device name list
+	//-------------------------------------------------------------------------------------------------
+	@Operation(summary = "Returns the system entries according to the query request")
+	@ApiResponses(value = {
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_OK, description = Constants.SWAGGER_HTTP_200_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = SystemListResponseDTO.class)) }),
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_OK, description = Constants.SWAGGER_HTTP_200_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = SystemListResponseDTO.class)) }),
```
rbocsi: These are the same. Need only one.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/ManagementAPI.java
```diff
@@ -178,15 +182,101 @@ public void removeDevices(final @RequestParam List<String> names) {
...
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) })
+	})
+	@PostMapping(path = ServiceRegistryConstants.HTTP_API_OP_SYSTEM_QUERY_PATH, consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
+	public @ResponseBody SystemListResponseDTO querySystems(@RequestBody(required = false) final SystemQueryRequestDTO dto,
+			@Parameter(
+					name =  "verbose",
+					description  = "Set true if you want the response to contain device details. (It should be configured in the Application properties as well.)",
+					example = "true")
+			@RequestParam final boolean verbose) {
+		logger.debug("querySystems started, verbose = " + Boolean.toString(verbose));
```
rbocsi: log4j does not like string concatenation, use instead parameters in the message:
```
logger.debug("querySystems started, verbose = {}", verbose);
```

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/repository/SystemRepository.java
```diff
@@ -1,10 +1,30 @@
...
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	public List<System> findAllByNameIn(final List<String> names);
+
+	//-------------------------------------------------------------------------------------------------
+	public Optional<System> findByName(final String name);
+
+	//-------------------------------------------------------------------------------------------------
+	public Page<System> findAllByNameIn(final List<String> names, Pageable pageble);
```
rbocsi: missing final

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+							.filter(s -> s.name().equals(systemEntity.getName()))
+							.findFirst()
+							.get()
+							.metadata();
+
+					final String version = toUpdate.stream()
+							.filter(s -> s.name().equals(systemEntity.getName()))
+							.findFirst()
+							.get()
+							.version();
```
rbocsi: You do a very similar thing twice. I think you should just find the matching SystemRequestDTO once (ends the expression at the get()), and use its metadata() and version() methods on the same object.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+	private static final Object LOCK = new Object();
+
 	private final Logger logger = LogManager.getLogger(this.getClass());
 
 	//=================================================================================================
 	// methods
 
+	//-------------------------------------------------------------------------------------------------
+	@Transactional(rollbackFor = ArrowheadException.class)
+	public SystemListResponseDTO createBulk(final List<SystemRequestDTO> candidates) {
```
rbocsi: This one should not return DTO, but some data structure that contains the necessary entities, see in DeviceDbService: ` List<Entry<Device, List<DeviceAddress>>>`

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+			final List<SystemAddress> systemAddressEntities = createSystemAddressEntities(candidates, systemEntities);
+			systemAddressRepo.saveAllAndFlush(systemAddressEntities);
+
+			//writing the device-system connections to the database
+			final List<DeviceSystemConnector> deviceSystemConnectorEntities = createDeviceSystemConnectorEntities(candidates);
+			if (deviceSystemConnectorEntities != null) {
+				deviceSystemConnectorRepo.saveAllAndFlush(deviceSystemConnectorEntities);
+			}
+
+			return createSystemListResponseDTOs(systemEntities, systemEntities.size());
```
rbocsi: The conversion to dto should be in the service layer

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+			logger.error(ex.getMessage());
+			logger.debug(ex);
+			throw new InternalServerError("Database operation error");
+		}
+
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	@Transactional(rollbackFor = ArrowheadException.class)
+	public SystemListResponseDTO updateBulk(final List<SystemRequestDTO> toUpdate) {
```
rbocsi: This one should not return DTO, but some data structure that contains the necessary entities, see in DeviceDbService: ` List<Entry<Device, List<DeviceAddress>>>`

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+
+					connection.setDevice(device);
+				}
+
+				//writing the changes into the database
+				deviceSystemConnectorRepo.deleteAll(connectionsToDelete);
+				deviceSystemConnectorRepo.saveAll(connectionsToUpdate);
+				deviceSystemConnectorRepo.flush();
+
+				return createSystemListResponseDTOs(systemEntities, systemEntities.size());
```
rbocsi: The conversion to dto should be in the service layer

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+
+		} catch (final Exception ex) {
+			logger.error(ex.getMessage());
+			logger.debug(ex);
+			throw new InternalServerError("Database operation error");
+		}
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public Optional<SystemRequestDTO> getByName(final String name) {
```
rbocsi: same as before. Do not return DTO. Even if we'd use DTOs here, it would definetely not be a requestDTO

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+		logger.debug("getByName started");
+		Assert.isTrue(!Utilities.isEmpty(name), "system name is missing or empty");
+
+		synchronized (LOCK) {
```
rbocsi, line 231: we don't want any synchronized in a getter

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+	//-------------------------------------------------------------------------------------------------
+	public Optional<System> getSystemByName(final String name) {
+		logger.debug("getSystemByName started");
+		Assert.isTrue(!Utilities.isEmpty(name), "system name is missing or empty");
+
+		return systemRepo.findByName(name);
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public SystemResponseDTO createSystemResponseDTO(final System system) {
```
rbocsi: this shouldn't be here but in the service layer with a parameter list that matches the new output of getByName().

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+					system.getUpdatedAt().toString()
+					);
+
+			return systemResponseDTO;
+		}
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	@SuppressWarnings("unchecked")
+	public SystemListResponseDTO getPageByFilters(final SystemQueryRequestDTO dto, final String origin) {
```
rbocsi: returns a Page (see example in DeviceDbService). Input parameters should contain a PageRequest, don't do it here but in the service layer

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
 public class SystemDbService {
 
 	//=================================================================================================
 	// members
 
+	@Autowired
+	private SystemRepository systemRepo;
+
+	@Autowired
+	private PageService pageService;
```
rbocsi: This one shouldn't be used here but in the service layer.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+			}
+
+			SystemListResponseDTO result = null;
+
+			// with filters
+			if (systemEntries == null) {
+
+				final List<String> matchings = new ArrayList<>();
+
+				synchronized (LOCK) {
```
rbocsi, line 340: no synchronized in getter

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+							if (!metaDataMatch) {
+								continue;
+							}
+						}
+						if (!Utilities.isEmpty(dto.versions()) && !systemRepo.findAllByVersionIn(dto.versions()).contains(system)) {
+							continue;
+						}
+						if (!Utilities.isEmpty(dto.deviceNames()) && !dto.deviceNames().contains(deviceSystemConnectorRepo.findBySystem(system).get().getDevice().getName())) {
+							continue;
+						}
```
rbocsi: The ordering of the ifs is not optimal. You should use the following order (from simple to difficult):

- address type
- addresses
- version
- device names
- metadata

Metadata should be the last because that one is difficult and if we can filter out a system before it means we don't have to check the metadata requirements for that system.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	//checks if device name already exists, throws exception if it does not
+	private void checkDeviceNamesExist(final List<SystemRequestDTO> candidates) {
+		final List<String> candidateDeviceNames = candidates.stream()
+				.map(c -> c.deviceName())
+				.collect(Collectors.toList())
+				.stream()
+				.distinct()
```
rbocsi: maybe used the distinct() before the first collect, so you don't need the second stream() and collect().

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+				.collect(Collectors.toList());
+
+		final List<String> existingDeviceNames = deviceRepo.findAllByNameIn(candidateDeviceNames).stream()
+				.map(e -> e.getName())
+				.collect(Collectors.toList());
+
+		final List<String> notExistingDeviceNames = new ArrayList<>();
+		for (final String candidateDeviceName : candidateDeviceNames) {
+			if (!existingDeviceNames.contains(candidateDeviceName) && candidateDeviceName != null) {
```
rbocsi: why do you need here the candidateDeviceName != null expression?

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+				notExistingDeviceNames.add(candidateDeviceName);
+			}
+		}
+		if (notExistingDeviceNames.size() != 0) {
```
rbocsi: !notExistingDeviceNames.isEmpty() should be better

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+	//-------------------------------------------------------------------------------------------------
+	private final List<DeviceSystemConnector> createDeviceSystemConnectorEntities(final List<SystemRequestDTO> candidates) {
+		synchronized (LOCK) {
+
+			final List<DeviceSystemConnector> connections = new ArrayList<>();
+
+			for (final SystemRequestDTO candidate : candidates) {
+				if (candidate.deviceName() != null) {
+					final Device device = deviceRepo.findAllByNameIn(Arrays.asList(candidate.deviceName())).getFirst();
+					final System system = systemRepo.findAllByNameIn(Arrays.asList(candidate.name())).getFirst();
```
rbocsi: just write a findByName method into the appropriate repositories

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -1,18 +1,588 @@
...
+					connections.add(new DeviceSystemConnector(device, system));
+				}
+			}
+
+			return connections;
+		}
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	private SystemListResponseDTO createSystemListResponseDTOs(final List<System> systems, final long pageSize) {
```
rbocsi: DTO creation should be in the service layer

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -53,11 +61,14 @@ public class ManagementService {
 	private ServiceDefinitionDbService serviceDefinitionDbService;
 
 	@Autowired
-	private ManagementNormalization normalizer;
+	private SystemDbService systemDbService;
 
 	@Autowired
 	private DTOConverter dtoConverter;
 
+    @Value("${service.discovery.verbose}")
```
rbocsi: this one should be come from a constant.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -160,15 +155,108 @@ public ServiceDefinitionListResponseDTO createServiceDefinitions(final ServiceDe
...
+			throw new InternalServerError(ex.getMessage(), origin);
+
+		}
+
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public SystemListResponseDTO querySystems(final SystemQueryRequestDTO dto, final boolean verbose, final String origin) {
+
+		logger.debug("querySystems started, verbose = " + verbose);
```
rbocsi: use message parameters here

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -160,15 +155,108 @@ public ServiceDefinitionListResponseDTO createServiceDefinitions(final ServiceDe
...
+	//-------------------------------------------------------------------------------------------------
+	public SystemListResponseDTO createSystems(final SystemListRequestDTO dto, final String origin) {
+		logger.debug("createSystems started");
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final List<SystemRequestDTO> normalized = validator.validateAndNormalizeCreateSystems(dto, origin);
+
+		try {
+
+			return systemDbService.createBulk(normalized);
```
rbocsi: DTO conversion should be here

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -160,15 +155,108 @@ public ServiceDefinitionListResponseDTO createServiceDefinitions(final ServiceDe
...
+
+		logger.debug("querySystems started, verbose = " + verbose);
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final SystemQueryRequestDTO normalized = validator.validateAndNormalizeQuerySystems(dto, origin);
+
 		try {
-			final List<ServiceDefinition> entities = serviceDefinitionDbService.createBulk(normalizedNames);
-			return dtoConverter.convertServiceDefinitionEntityListToDTO(entities);
+			final SystemListResponseDTO result = systemDbService.getPageByFilters(normalized, origin);
```
rbocsi: DTO conversion should be here. It is also has considered the verbose requirement

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -160,15 +155,108 @@ public ServiceDefinitionListResponseDTO createServiceDefinitions(final ServiceDe
...
+
+			//we do not provide device information (except for the name), if the verbose mode is not enabled, or the user set it false in the query param
+			if (!verbose || !verboseEnabled) {
+				final List<SystemResponseDTO> resultTerse = new ArrayList<>();
```
rbocsi, line 200: I must be stupid, but I don't know what resultTerse means.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -160,15 +155,108 @@ public ServiceDefinitionListResponseDTO createServiceDefinitions(final ServiceDe
...
+	//-------------------------------------------------------------------------------------------------
+	public SystemListResponseDTO updateSystems(final SystemListRequestDTO dto, final String origin) {
+
+		logger.debug("updateSystems started");
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final List<SystemRequestDTO> normalized = validator.validateAndNormalizeUpdateSystems(dto, origin);
+
+		try {
+			return systemDbService.updateBulk(normalized);
```
rbocsi: DTO conversion should be here

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/SystemDiscoveryService.java
```diff
@@ -0,0 +1,190 @@
...
+	@Autowired
+	private SystemDiscoveryValidation validator;
+
+	@Autowired
+	private AddressMatching addressMatcher;
+
+	@Autowired
+	private SystemDbService dbService;
+
+    @Value("${service.discovery.verbose}")
```
rbocsi: use constant

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/SystemDiscoveryService.java
```diff
@@ -0,0 +1,190 @@
...
+    private boolean verboseEnabled;
+
+	private final Logger logger = LogManager.getLogger(this.getClass());
+
+	//=================================================================================================
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	public Entry<SystemResponseDTO, Boolean> registerSystem(final SystemRequestDTO dto, final String origin) {
```
rbocsi, line 68: db services should not return dto, dto conversion should be here

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/SystemDiscoveryService.java
```diff
@@ -0,0 +1,190 @@
...
+					new SystemQueryRequestDTO(
+							null,
+							normalized.systemNames(),
+							normalized.addresses(),
+							normalized.addressType(),
+							normalized.metadataRequirementList(),
+							normalized.versions(),
+							normalized.deviceNames()),
+					origin);
+
```
rbocsi, line 121: db services should not return dto, dto conversion should be here (that handles the verbose requirement as well)

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/matching/AddressMatching.java
```diff
@@ -0,0 +1,58 @@
...
+	//-------------------------------------------------------------------------------------------------
+	// checks if the two list contain exactly the same addresses (the order does not matter)
+	public boolean isAddressListMatching(final List<AddressDTO> addresses1, final List<AddressDTO> addresses2) {
+
+		if (addresses1.size() != addresses2.size()) {
+			return false;
+		}
+
+		List<AddressDTO> temp1 = new ArrayList<>(addresses1);
+		final List<AddressDTO> temp2 = new ArrayList<>(addresses2);
```
rbocsi: temp2 is unnecessary since you don't modify addresses2 list

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/matching/AddressMatching.java
```diff
@@ -0,0 +1,58 @@
...
+
+		int temp1Size = temp1.size();
+		while (temp1Size != 0) {
```
rbocsi: It is better if temp1Size > 0

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/matching/AddressMatching.java
```diff
@@ -0,0 +1,58 @@
...
+	// returns the original list, if there was no match
+	private List<AddressDTO> removeMatch(final List<AddressDTO> addresses, final AddressDTO required) {
+		for (final AddressDTO address : addresses) {
+			if (address.equals(required)) {
+				addresses.remove(address);
+				return addresses;
+			}
+		}
+		return addresses;
+	}
```
rbocsi: I don't understand why is this necessary. It seems to me that addresses.remove(required) does the same.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ManagementNormalization.java
```diff
@@ -12,24 +12,85 @@
...
+		return normalized;
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public SystemQueryRequestDTO normalizeSystemQueryRequestDTO(final SystemQueryRequestDTO dto) {
+		logger.debug("normalizeSystemQueryRequestDTO started");
+		Assert.notNull(dto, "SystemQueryRequestDTO is null");
+
+		return new SystemQueryRequestDTO(
+				dto.pagination(), //no need to normailze, because it will happen in the getPageRequest method
```
rbocsi: typo in comment normailze => normalize

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
```diff
@@ -169,4 +239,212 @@ public void validateCreateServiceDefinition(final ServiceDefinitionListRequestDT
...
+
+			if (Utilities.isEmpty(system.name())) {
+				throw new InvalidParameterException("System name is empty", origin);
+			}
+
+			if (names.contains(system.name())) {
+				throw new InvalidParameterException("Duplicate system name: " + system.name(), origin);
+			}
+
+			if (system.name().length() > ArrowheadEntity.VARCHAR_SMALL) {
```
rbocsi: should use Constants.SYSTEM_NAME_LENGTH instead

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
```diff
@@ -169,4 +239,212 @@ public void validateCreateServiceDefinition(final ServiceDefinitionListRequestDT
...
+		validateCreateSystems(dto, origin);
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public void validateQuerySystems(final SystemQueryRequestDTO dto, final String origin) {
+		logger.debug("validateQuerySystems started");
+
+		if (dto == null) {
+			throw new InvalidParameterException("Request payload is missing", origin);
+		}
```
rbocsi: I think it is not a problem if the dto is null in this case (the similar method about devices does not throw an exception).

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
```diff
@@ -169,4 +239,212 @@ public void validateCreateServiceDefinition(final ServiceDefinitionListRequestDT
...
+		logger.debug("validateAndNormalizeUpdateSystems started");
+
+		return validateAndNormalizeCreateSystems(dto, origin);
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public SystemQueryRequestDTO validateAndNormalizeQuerySystems(final SystemQueryRequestDTO dto, final String origin) {
+		logger.debug("validateAndNormalizeQuerySystems started");
+
+		validateQuerySystems(dto, origin);
```
rbocsi, line 391: In this method you have to accept null dto as well (maybe returns an empty dto in this case without any filter)

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/SystemDiscoveryValidation.java
```diff
@@ -0,0 +1,176 @@
...
+
+		return normalized;
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public SystemLookupRequestDTO validateAndNormalizeLookupSystem(final SystemLookupRequestDTO dto, final String origin) {
+
+		validateLookupSystem(dto, origin);
+
+		final SystemLookupRequestDTO normalized = normalizer.normalizeSystemLookupRequestDTO(dto);
```
rbocsi, line 141: in this method you can have a null dto. In this case you may return an empty dto without any filter

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/version/VersionValidator.java
```diff
@@ -0,0 +1,59 @@
...
+	private static final int NUMBER_OF_SEPARATORS = 2;
+
+
+	//=================================================================================================
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	public void validateNormalizedVersion(final String version) {
+
+		logger.debug("Validate version started: " + version);
```
rbocsi: use message parameters

rbocsi (changes requested):
see inline comments

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
Katinka666: I don't think it would make the same result, because the same device can be associated to different systems, and the device-system connections will be distinguished even if the device is the same.

rbocsi: I think it is the same result, that is my point. It is the same result with less operations.

Katinka666: That is true. Sorry, I was looking at a different function.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/repository/DeviceSystemConnectorRepository.java
```diff
@@ -15,4 +16,7 @@ public interface DeviceSystemConnectorRepository extends RefreshableRepository<D
 
 	//-------------------------------------------------------------------------------------------------
 	public Optional<DeviceSystemConnector> findBySystem(final System system);
+
+	//-------------------------------------------------------------------------------------------------
+	public Optional<List<DeviceSystemConnector>> findBySystemIn(final List<System> system);
```
rbocsi: Do you need Optional here? In other cases we just use List return value, and in worst case it returns an empty list.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
-					.map(sa -> new AddressDTO(sa.getAddressType().toString(), sa.getAddress()))
-					.collect(Collectors.toList());
-
-			final Optional<DeviceSystemConnector> deviceSystemConncetor = deviceSystemConnectorRepo.findBySystem(system);
-
-			DeviceResponseDTO deviceResponseDTO = null;
-
-			if (deviceSystemConncetor.isPresent()) {
+			// device addresses
+			final List<DeviceAddress> deviceAddresses = device == null ? new ArrayList<DeviceAddress>() : deviceAddressRepo.findAllByDevice(device);
```
rbocsi: use `List.of()` instead of `new ArrayList<DeviceAddress>()`

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
-					&& Utilities.isEmpty(dto.addressType())
-					&& Utilities.isEmpty(dto.metadataRequirementList())
-					&& Utilities.isEmpty(dto.versions())
-					&& Utilities.isEmpty(dto.deviceNames())) {
-
-				if (pageRequest != null) {
-					systemEntries = systemRepo.findAll(pageRequest);
+		synchronized (LOCK) {
+			try {
+				Page<System> systemEntries = new PageImpl<System>(new ArrayList<System>());
```
rbocsi: too much System, use diamonds when possible:

`Page<System> systemEntries = new PageImpl<>(new ArrayList<>());`

Edit: After I  saw the whole code here you don't need right side at all, just

`Page<System> systemEntries = null;`

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
-						if (!Utilities.isEmpty(dto.metadataRequirementList())) {
-							final Map<String, Object> metadata = Utilities.fromJson(system.getMetadata(), new TypeReference<Map<String, Object>>() { });
-							boolean metaDataMatch = false;
-							for (final MetadataRequirementDTO requirement : dto.metadataRequirementList()) {
-								if (MetadataRequirementsMatcher.isMetadataMatch(metadata, requirement)) {
-									metaDataMatch = true;
-									break;
-								}
+							// version
+							if (!Utilities.isEmpty(versions) && !systemRepo.findAllByVersionIn(versions).contains(system)) {
```
rbocsi: I didn't notice the first time, but: the contains() only works well if the System.equals() is right. Since we don't override the Object.equals() it returns true only if the two systems are the same object, which may be not the case. You have to override equals() in System in a way that two system is equals if their db id is the same.

You can use the Eclipse's built in option to generate equals (and hashCode), it is in the Source menu. If you use that please note that you have to manually reformat the result, because the generated code is not clean enough (no finals, missing parantheses in ifs, etc).

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
+								continue;
 							}
-							if (!metaDataMatch) {
+
+							// device names
+							if (!Utilities.isEmpty(deviceNames) && deviceSystemConnectorRepo.findBySystem(system).isPresent()
+									&& !deviceNames.contains(deviceSystemConnectorRepo.findBySystem(system).get().getDevice().getName())) {
```
rbocsi, line 290: I don't like here that you call the same db operation twice. Refactor this to something like that:
```
if (!Utilities.isEmpty(deviceNames)) {
    // do the deviceSystemConnectorRepo.findBySystem(system) once and assign to a local variable
   if ( /* checking optional and not contains here */ ) {
       continue;
  }
}
```

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
-							continue;
-						}
-						if (!Utilities.isEmpty(dto.deviceNames()) && !dto.deviceNames().contains(deviceSystemConnectorRepo.findBySystem(system).get().getDevice().getName())) {
-							continue;
+
+							// metadata
+							if (!Utilities.isEmpty(metadataRequirementList)) {
+								boolean metaDataMatch = false;
+								for (final Map.Entry<String, Object> requirement : metadataRequirementList.entrySet()) {
+									if (MetadataRequirementsMatcher.isMetadataMatch(Utilities.fromJson(system.getMetadata(), new TypeReference<Map<String, Object>>() { }), (MetadataRequirementDTO) requirement)) {
```
rbocsi: `Utilities.fromJson(system.getMetadata(), new TypeReference<Map<String, Object>>() { })` should be do once before the loop

after the previous fixes, you don't have to cast the requirement

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
 		}
+
 	}
 
 	//-------------------------------------------------------------------------------------------------
-	@SuppressWarnings("unchecked")
-	public SystemListResponseDTO getPageByFilters(final SystemQueryRequestDTO dto, final String origin) {
+	public Page<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> getPageByFilters(
+			final PageRequest pagination, final List<String> systemNames, final List<String> addresses, final AddressType addressType,
+			final Map<String, Object> metadataRequirementList, final List<String> versions, final List<String> deviceNames) {
```
rbocsi: wrong parameter here: should be final List<MetadataRequirementDTO> metadataRequirementList

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
-						if (!Utilities.isEmpty(dto.versions()) && !systemRepo.findAllByVersionIn(dto.versions()).contains(system)) {
-							continue;
-						}
-						if (!Utilities.isEmpty(dto.deviceNames()) && !dto.deviceNames().contains(deviceSystemConnectorRepo.findBySystem(system).get().getDevice().getName())) {
-							continue;
+
+							// metadata
+							if (!Utilities.isEmpty(metadataRequirementList)) {
+								boolean metaDataMatch = false;
+								for (final Map.Entry<String, Object> requirement : metadataRequirementList.entrySet()) {
```
rbocsi: wrong loop because of wrong parameter before: should be `for (final MetadataRequirementDTO requirement : metadataRequirementList) {`

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
-						}
-						if (!Utilities.isEmpty(dto.versions()) && !systemRepo.findAllByVersionIn(dto.versions()).contains(system)) {
-							continue;
-						}
-						if (!Utilities.isEmpty(dto.deviceNames()) && !dto.deviceNames().contains(deviceSystemConnectorRepo.findBySystem(system).get().getDevice().getName())) {
-							continue;
+
+							// metadata
+							if (!Utilities.isEmpty(metadataRequirementList)) {
+								boolean metaDataMatch = false;
```
rbocsi: metaData => metadata everywhere

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
-				if (pageRequest != null) {
-					systemEntries = systemRepo.findAllByNameIn(matchings, pageRequest);
-					result = createSystemListResponseDTOs(systemEntries.toList(), systemEntries.getTotalElements());
-				} else {
-					final List<System> systemEntriesList = systemRepo.findAllByNameIn(matchings);
-					result = createSystemListResponseDTOs(systemEntriesList, systemEntriesList.size());
-				}
-			}
+				// system addresses
+				final List<SystemAddress> systemAddresses = systemAddressRepo.findAllBySystemIn(systemEntries.toList());
```
rbocsi: It is better to use getContent() here instead of toList() because the latter one always create a new list (and you use it 3 times), while the former is not.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -224,171 +211,124 @@ public void deleteByNameList(final List<String> names) {
...
 
-			return result;
+				// device-system connections
+				final Optional<List<DeviceSystemConnector>> deviceSystemConnections = deviceSystemConnectorRepo.findBySystemIn(systemEntries.toList());
 
-		} catch (final Exception ex) {
+				final List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> result = createTriples(systemEntries.toList(), systemAddresses, deviceSystemConnections.isEmpty()
+						? new ArrayList<DeviceSystemConnector>() : deviceSystemConnections.get());
```
rbocsi: use List.of() as first option here

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -464,24 +404,26 @@ private void checkSystemNamesExist(final List<SystemRequestDTO> requested) {
 	//-------------------------------------------------------------------------------------------------
 	//checks if device name already exists, throws exception if it does not
 	private void checkDeviceNamesExist(final List<SystemRequestDTO> candidates) {
-		final List<String> candidateDeviceNames = candidates.stream()
-				.map(c -> c.deviceName())
-				.collect(Collectors.toList())
-				.stream()
-				.distinct()
-				.collect(Collectors.toList());
+
```
rbocsi, line 414: I think we misunderstood each other here. I have no problem with using lambda expression here, it is just not optimal (and also not handle null values).

```
final List<String> candidateDeviceNames = candidates.stream()
                                .filter(c -> c.deviceName() != null)
				.map(c -> c.deviceName())
				.distinct()
				.collect(Collectors.toList());
```

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -537,52 +479,35 @@ private final List<DeviceSystemConnector> createDeviceSystemConnectorEntities(fi
...
 
 	//-------------------------------------------------------------------------------------------------
-	private SystemListResponseDTO createSystemListResponseDTOs(final List<System> systems, final long pageSize) {
-		final List<SystemResponseDTO> result = new ArrayList<>();
+	private List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> createTriples(
+			final List<System> systems, final List<SystemAddress> addresses, final List<DeviceSystemConnector> deviceConnections) {
+
+		Assert.notNull(systems, "systems are null");
+
+		final List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> result = new ArrayList<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>>();
```
rbocsi: on the right side you can use just a diamond: `new ArrayList<>()`.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -537,52 +479,35 @@ private final List<DeviceSystemConnector> createDeviceSystemConnectorEntities(fi
...
-
-					 deviceResponseDTO = new DeviceResponseDTO(
-							device.getName(),
-							Utilities.fromJson(device.getMetadata(), new TypeReference<Map<String, Object>>() { }),
-							deviceAddresses,
-							device.getCreatedAt().toString(),
-							device.getUpdatedAt().toString());
-				}
+			// corresponding device addresses
+			final List<DeviceAddress> deviceAddressList = device == null ? new ArrayList<DeviceAddress>() : deviceAddressRepo.findAllByDevice(device);
```
rbocsi: List.of() for creating an empty list

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -67,7 +71,7 @@ public class ManagementService {
 	private DTOConverter dtoConverter;
 
     @Value("${service.discovery.verbose}")
-    private boolean verboseEnabled;
+    private static boolean verboseEnabled;
```
rbocsi: We misunderstand each other here. This member is not static, it is just in the ˙@Value˙ annotation I don't want to see string literals. '${service.discovery.verbose}' should be a constant in the ServiceRegistryConstants class (already have examples there) and here you just use the name of the constant.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -170,7 +174,8 @@ public SystemListResponseDTO createSystems(final SystemListRequestDTO dto, final
 
 		try {
 
-			return systemDbService.createBulk(normalized);
+			final List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> entities = systemDbService.createBulk(normalized);
+			return dtoConverter.convertSystemTripleListToDTO(entities);
```
rbocsi: I'd rename this method to convertSystemTripletListToDTO.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -187,39 +192,31 @@ public SystemListResponseDTO createSystems(final SystemListRequestDTO dto, final
...
 		try {
-			final SystemListResponseDTO result = systemDbService.getPageByFilters(normalized, origin);
-
+			final PageRequest pageRequest = pageService.getPageRequest(normalized.pagination(), Direction.DESC, System.SORTABLE_FIELDS_BY, System.DEFAULT_SORT_FIELD, origin);
+			final Page<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> page = systemDbService.getPageByFilters(
+					pageRequest,
+					normalized.systemNames(),
+					normalized.addresses(),
+					Utilities.isEmpty(normalized.addressType()) ? null : AddressType.valueOf(normalized.addressType()),
+					Utilities.isEmpty(normalized.metadataRequirementList()) ? new HashMap<String, Object>()
```
rbocsi: It seems to me you have overthinked this one. You just have to pass the normalized.metadataRequirementList() here. If it is null or empty list the dbService handles it.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -187,39 +192,31 @@ public SystemListResponseDTO createSystems(final SystemListRequestDTO dto, final
...
+							: Utilities.fromJson(Utilities.toJson(normalized.metadataRequirementList()), new TypeReference<Map<String, Object>>() { }),
+					normalized.versions(),
+					normalized.deviceNames());
+
+			final SystemListResponseDTO result = dtoConverter.convertSystemTriplesToDTO(page);
```
rbocsi: I'd rename this method to convertSystemTripletPageToDTO.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -187,39 +192,31 @@ public SystemListResponseDTO createSystems(final SystemListRequestDTO dto, final
...
-							systemResponseDTO.createdAt(),
-							systemResponseDTO.updatedAt()
-							));
-				}
-
-				return new SystemListResponseDTO(resultTerse, result.count());
+				return dtoConverter.convertSystemListResponseDtoToTerse(result);
 			}
 
+			java.lang.System.out.println(result.count());
```
rbocsi: remove this line

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/SystemDiscoveryService.java
```diff
@@ -42,8 +53,11 @@ public class SystemDiscoveryService {
 	@Autowired
 	private SystemDbService dbService;
 
+	@Autowired
+	private DTOConverter dtoConverter;
+
     @Value("${service.discovery.verbose}")
-    private boolean verboseEnabled;
+    private static boolean verboseEnabled;
```
rbocsi: See my appropriate comment in the ManagementService class.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/SystemDiscoveryService.java
```diff
@@ -59,24 +73,23 @@ public Entry<SystemResponseDTO, Boolean> registerSystem(final SystemRequestDTO d
...
+				final Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>> existingSystem = optional.get();
 
 				// We should check if every property is the same
-				checkSameSystemAttributes(existingSystemAsDTO, normalized);
+				checkSameSystemAttributes(existingSystem, normalized);
 
 				// Convert to response and return
-				final System existing = dbService.getSystemByName(existingSystemAsDTO.name()).get();
-				final SystemResponseDTO existingSystemAsResponseDTO = dbService.createSystemResponseDTO(existing);
+				final SystemResponseDTO existingSystemAsResponseDTO = dtoConverter.convertSystemTripleToDTO(existingSystem);
```
rbocsi: I'd rename this method to convertSystemTripletToDTO

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/SystemDiscoveryService.java
```diff
@@ -94,40 +107,22 @@ public SystemListResponseDTO lookupSystem(final SystemLookupRequestDTO dto, fina
...
-							normalized.deviceNames()),
-					origin);
+			final PageRequest pageRequest = PageRequest.of(0, Integer.MAX_VALUE, Direction.DESC, System.DEFAULT_SORT_FIELD);
+			final Page<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> page = dbService.getPageByFilters(
+					pageRequest,
+					normalized.systemNames(),
+					normalized.addresses(),
+					Utilities.isEmpty(normalized.addressType()) ? null : AddressType.valueOf(normalized.addressType()),
+					Utilities.isEmpty(normalized.metadataRequirementList()) ? new HashMap<String, Object>()
+							: Utilities.fromJson(Utilities.toJson(normalized.metadataRequirementList()), new TypeReference<Map<String, Object>>() { }),
```
rbocsi: see my relevant comment in ManagementService where you use getPageByFilters()

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/SystemDiscoveryService.java
```diff
@@ -155,36 +150,41 @@ public boolean revokeSystem(final String name, final String origin) {
...
 		logger.debug("checkSameSystemAttributes started");
+		Assert.isTrue(existing.getLeft().getName().equals(dto.name()), "The systems are not identical!");
 
-		Assert.isTrue(existing.name().equals(dto.name()), "The systems are not identical!");
+		final System existingSystem = existing.getLeft();
 
 		// metadata
-		if (!existing.metadata().equals(dto.metadata())) {
-			throw new InvalidParameterException("System with name: " + existing.name() + " already exists, but provided metadata is not matching");
+		if (!existingSystem.getMetadata().equals(Utilities.toJson(dto.metadata()))) {
```
rbocsi: this is wrong. The two metadata can be the same even if their string representation is not the same. You have to transform the string to Map<String, Object> and compare the two maps, not the other way.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/dto/DTOConverter.java
```diff
@@ -80,4 +86,89 @@ public ServiceDefinitionResponseDTO convertServiceDefinitionEntityToDTO(final Se
 
 		return new ServiceDefinitionResponseDTO(entity.getName(), Utilities.convertZonedDateTimeToUTCString(entity.getCreatedAt()), Utilities.convertZonedDateTimeToUTCString(entity.getUpdatedAt()));
 	}
+
+	//-------------------------------------------------------------------------------------------------
+	public SystemListResponseDTO convertSystemTriplesToDTO(final Page<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> entities) {
+		logger.debug("convertSystemTriplesToDTO started...");
+
+		final List<SystemResponseDTO> result = new ArrayList<>();
```
rbocsi: You can initialize the capacity of the ArrayList with entities.getSize()

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/dto/DTOConverter.java
```diff
@@ -80,4 +86,89 @@ public ServiceDefinitionResponseDTO convertServiceDefinitionEntityToDTO(final Se
...
+		}
+
+		return new SystemListResponseDTO(result, entities.getTotalElements());
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public SystemListResponseDTO convertSystemTripleListToDTO(final List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> entities) {
+		logger.debug("convertSystemTriplesToDTO started...");
+
+		final List<SystemResponseDTO> result = new ArrayList<>();
```
rbocsi: You can initialize the capacity of the ArrayList with entities.getSize()

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/dto/DTOConverter.java
```diff
@@ -80,4 +86,89 @@ public ServiceDefinitionResponseDTO convertServiceDefinitionEntityToDTO(final Se
...
+						Utilities.convertZonedDateTimeToUTCString(device.getCreatedAt()),
+						Utilities.convertZonedDateTimeToUTCString(device.getUpdatedAt())),
+				Utilities.convertZonedDateTimeToUTCString(system.getCreatedAt()),
+				Utilities.convertZonedDateTimeToUTCString(system.getUpdatedAt()));
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public SystemListResponseDTO convertSystemListResponseDtoToTerse(final SystemListResponseDTO verbose) {
+
+		final List<SystemResponseDTO> terse = new ArrayList<SystemResponseDTO>();
```
rbocsi: You can initialize the capacity of the ArrayList with verbose.entries().getSize()

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ManagementNormalization.java
```diff
@@ -62,10 +66,20 @@ public List<SystemRequestDTO> normalizeSystemRequestDTOs(final SystemListRequest
...
 		logger.debug("normalizeSystemQueryRequestDTO started");
-		Assert.notNull(dto, "SystemQueryRequestDTO is null");
+
+		if (dto == null) {
+			return new SystemQueryRequestDTO(
+					new PageDTO(0, Integer.MAX_VALUE, Direction.DESC.toString(), System.DEFAULT_SORT_FIELD),
+					null,
+					null,
+					null,
+					new ArrayList<MetadataRequirementDTO>(),
```
rbocsi: List.of() for empty list, but I think that null should suffice as well.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/SystemDiscoveryNormalization.java
```diff
@@ -52,7 +53,10 @@ public SystemRequestDTO normalizeSystemRequestDTO(final SystemRequestDTO dto) {
 	//-------------------------------------------------------------------------------------------------
 	public SystemLookupRequestDTO normalizeSystemLookupRequestDTO(final SystemLookupRequestDTO dto) {
 		logger.debug("normalizeSystemLookupRequestDTO started");
-		Assert.notNull(dto, "SystemLookupRequestDTO is null");
+
+		if (dto == null) {
+			return new SystemLookupRequestDTO(null, null, null, new ArrayList<MetadataRequirementDTO>(), null, null);
```
rbocsi: List.of() for empty list, but I think that null should suffice as well.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
```diff
@@ -320,34 +319,37 @@ public void validateUpdateSystems(final SystemListRequestDTO dto, final String o
...
+			if (!Utilities.isEmpty(dto.addressType()) && !Utilities.isEnumValue(dto.addressType(), AddressType.class)) {
+				throw new InvalidParameterException("Invalid address type: " + dto.addressType(), origin);
+			}
 
-		if (!Utilities.isEmpty(dto.metadataRequirementList()) && Utilities.containsNull(dto.metadataRequirementList())) {
-			throw new InvalidParameterException("Metadata requirement list contains null element", origin);
-		}
+			if (!Utilities.isEmpty(dto.addresses()) && Utilities.isEmpty(dto.addressType())) {
+				throw new InvalidParameterException("Address list is not empty, but the address type was not specified!");
+			}
```
rbocsi: I think it should be possible to just filtering some addresses without specifying the type.

rbocsi (changes requested):
see inline comments. I found new problems.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
Katinka666: And is it ok if I pull it from the application.properties file in ServiceRegistryConstans?

rbocsi: No, you use the annotation here, you just create a constants there.

Here is an example:
```
	public static final String ALLOW_SELF_ADDRESSING = "allow.self.addressing";
	public static final String $ALLOW_SELF_ADDRESSING_WD = "${" + ALLOW_SELF_ADDRESSING + ":true}";
```

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
Katinka666: Can I use the following, since the system name is unique?
`systemRepo.findAllByVersionIn(versions).stream().map(s -> s.getName()).collect(Collectors.toList()).contains(system.getName())`

Katinka666:
@rbocsi please review again

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
rbocsi: Yes, it could work. But if you want to use system names there is a more concise version of it:
```
systemRepo.findAllByVersionIn(versions).stream().filter(s -> s.getName().equals(system.getName())).findAny().isPresent()
```

and you can remove the not operator if you change the isPresent() to isEmpty() in the expression.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/ServiceRegistryConstants.java
```diff
@@ -36,6 +36,8 @@ public final class ServiceRegistryConstants {
 	public static final String $ALLOW_SELF_ADDRESSING_WD = "${" + ALLOW_SELF_ADDRESSING + ":true}";
 	public static final String ALLOW_NON_ROUTABLE_ADDRESSING = "allow.non.routable.addressing";
 	public static final String $ALLOW_NON_ROUTABLE_ADDRESSING_WD = "${" + ALLOW_NON_ROUTABLE_ADDRESSING + ":true}";
+	public static final String SERVICE_DISCOVERY_VERBOSE = "service.discovery.verbose";
+	public static final String $SERVICE_DISCOVERY_VERBOSE_WD = "${" + SERVICE_DISCOVERY_VERBOSE + ":true}";
```
rbocsi: default should be false

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -291,16 +291,24 @@ public Page<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress
 								continue;
 							}
 
+							if (!Utilities.isEmpty(deviceNames)) {
+								Optional<DeviceSystemConnector> connection = deviceSystemConnectorRepo.findBySystem(system);
```
rbocsi: missing final

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -291,16 +291,24 @@ public Page<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress
...
+							}
+
 							// metadata
 							if (!Utilities.isEmpty(metadataRequirementList)) {
-								boolean metaDataMatch = false;
-								for (final Map.Entry<String, Object> requirement : metadataRequirementList.entrySet()) {
-									if (MetadataRequirementsMatcher.isMetadataMatch(Utilities.fromJson(system.getMetadata(), new TypeReference<Map<String, Object>>() { }), (MetadataRequirementDTO) requirement)) {
-										metaDataMatch = true;
+								boolean metadataMatch = false;
+								Map<String, Object> systemMetadata = Utilities.fromJson(system.getMetadata(), new TypeReference<Map<String, Object>>() { });
```
rbocsi: missing final

rbocsi (approved):
I made some new comments, but those are minor issues. After you fix them, no further review is needed from me.

borditamas (approved):

PR #21, opened 2024-09-29
Kj/config

Katinka666:


serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/ManagementAPI.java
```diff
@@ -88,7 +93,27 @@ public class ManagementAPI {
...
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_BAD_REQUEST, description = Constants.SWAGGER_HTTP_400_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) }),
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_UNAUTHORIZED, description = Constants.SWAGGER_HTTP_401_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) }),
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_FORBIDDEN, description = Constants.SWAGGER_HTTP_403_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) }),
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_INTERNAL_SERVER_ERROR, description = Constants.SWAGGER_HTTP_500_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) })
+	})
+	@PostMapping(path = ServiceRegistryConstants.HTTP_API_OP_GET_CONFIG_PATH, produces = MediaType.APPLICATION_JSON_VALUE)
```
rbocsi: It is a GET

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/ManagementAPI.java
```diff
@@ -88,7 +93,27 @@ public class ManagementAPI {
 
 	// CONFIG
 
-	// get-config GET /config(list of config keys as query params) -> Empty input list means all is required
+	//-------------------------------------------------------------------------------------------------
+	@Operation(summary = "Return a list of configurations")
+	@ApiResponses(value = {
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_OK, description = Constants.SWAGGER_HTTP_200_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = LogEntryListResponseDTO.class)) }),
```
rbocsi: It does not return a LogEntryListResponseDTO

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/ManagementAPI.java
```diff
@@ -88,7 +93,27 @@ public class ManagementAPI {
...
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_FORBIDDEN, description = Constants.SWAGGER_HTTP_403_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) }),
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_INTERNAL_SERVER_ERROR, description = Constants.SWAGGER_HTTP_500_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) })
+	})
+	@PostMapping(path = ServiceRegistryConstants.HTTP_API_OP_GET_CONFIG_PATH, produces = MediaType.APPLICATION_JSON_VALUE)
+	@ResponseBody public KeyValuesDTO getConfig(final @RequestParam(required = false) List<String> keys) {
+		logger.debug("getConfig started ...");
+
+		final String origin = HttpMethod.POST.name() + " " + ServiceRegistryConstants.HTTP_API_MANAGEMENT_PATH + ServiceRegistryConstants.HTTP_API_OP_GET_CONFIG_PATH;
```
rbocsi: it is a GET

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ConfigService.java
```diff
@@ -0,0 +1,89 @@
...
+	//Configuration
+
+	@Value(ServiceRegistryConstants.$ALLOW_SELF_ADDRESSING_WD)
+	private boolean allowSelfAddressing;
+
+	@Value(ServiceRegistryConstants.$ALLOW_NON_ROUTABLE_ADDRESSING_WD)
+	private boolean allowNonRoutableAddressing;
+
+	@Value(ServiceRegistryConstants.$SERVICE_DISCOVERY_VERBOSE_WD)
+	private boolean serviceDiscoveryVerbose;
```
rbocsi: I don't like the idea to extend this class whenever somebody needs an other property. Instead of this you should autowire the Environment bean that contains all the properties of the application.properties. You also need a set that contains the forbidden keys (db related keys, cert related key, everything we should not provide via the get-config endpoint).

Using the method we don't need to modify the code whenever a new property is introduced (only if the property is a secret one => in this case you have to extend the forbidden keys set).

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ConfigService.java
```diff
@@ -0,0 +1,89 @@
...
+
+		final Map<String, String> configKeyValues = Map.of(
+				ServiceRegistryConstants.ALLOW_SELF_ADDRESSING, String.valueOf(allowSelfAddressing),
+				ServiceRegistryConstants.ALLOW_NON_ROUTABLE_ADDRESSING, String.valueOf(allowNonRoutableAddressing),
+				ServiceRegistryConstants.SERVICE_DISCOVERY_VERBOSE, String.valueOf(serviceDiscoveryVerbose));
+
+		try {
+			final List<String> normalized = validator.validateAndNormalizeConfigKeyList(keys);
+
+			// Empty input list means all is required
```
rbocsi: I know we said that if there is no input list that means all is required, but I changed my mind. For security reasons, I think the caller has to know what it wants.  

Also, if a key is invalid (or forbidden) we should just omit from the response, not send error back.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ConfigNormalization.java
```diff
@@ -0,0 +1,35 @@
...
+	//=================================================================================================
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	public List<String> normalizeConfigKeyList(final List<String> keys) {
+		logger.debug("normalizeConfigKeyList started");
+
+		if (keys == null) {
+			return List.of();
+		}
```
rbocsi: Based on previous comment null or empty list should return an error. Not here but in the validation.

rbocsi (changes requested):
see inline comments

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/ServiceRegistryConstants.java
```diff
@@ -40,6 +42,19 @@ public final class ServiceRegistryConstants {
...
+	public static final List<String> FORBIDDEN_KEYS = List.of(
+			// database related
+			"spring.datasource.url",
+			"spring.datasource.username",
+			"spring.datasource.password",
+			"spring.datasource.driver-class-name",
+			"spring.jpa.hibernate.ddl-auto",
+			"spring.jpa.show-sql",
+			// cert related
+			"authenticator.secret.key");
```
rbocsi: there are more cert related keys (just not appearing currently in the application.properties, because we don't use the cert authentication):

```
server.ssl.key-store-type=PKCS12
server.ssl.key-store=classpath:certificates/sentinel-storage.p12
server.ssl.key-store-password=123456
server.ssl.key-alias=sentinel-storage.testcloud2.aitia.arrowhead.eu
server.ssl.key-password=123456
server.ssl.client-auth=need
server.ssl.trust-store-type=PKCS12
server.ssl.trust-store=classpath:certificates/truststore.p12
server.ssl.trust-store-password=123456

disable.hostname.verifier=false
```

rbocsi (approved):
Approved, but see my comment.

borditamas (approved):

PR #26, opened 2024-10-02
ServiceDiscovery implementation

borditamas:


serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/utils/ServiceLookupPreprocessor.java
```diff
@@ -0,0 +1,32 @@
...
+	// members
+
+	private final Logger logger = LogManager.getLogger(this.getClass());
+
+	//=================================================================================================
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	public boolean isRestricted(final HttpServletRequest request) {
+		logger.debug("process started");
```
rbocsi: this method's name is not process

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/utils/ServiceLookupPreprocessor.java
```diff
@@ -0,0 +1,32 @@
...
+
+		if (request == null) {
+			return false;
```
rbocsi: Not really matters (request never will be null), but in this case the better option is to return true.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/entity/ServiceInstance.java
```diff
@@ -18,9 +16,8 @@ public class ServiceInstance extends ArrowheadEntity {
 	//=================================================================================================
 	// members
 
-	@Id
-	@GeneratedValue(strategy = GenerationType.IDENTITY)
-	private long id;
+	public static final List<String> SORTABLE_FIELDS_BY = List.of("id", "serviceInstanceId", "createdAt");
+	public static final String DEFAULT_SORT_FIELD = "createdAt";
```
rbocsi: I think service instance id would be a better default. It would be sorted basically system name and service definition name and version alphabetically.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -1,18 +1,443 @@
...
+					}
+
+					interfaceTemplateCache.put(
+							interfaceCandidate.templateName(),
+							serviceInterfaceTemplateRepo.saveAndFlush(new ServiceInterfaceTemplate(interfaceCandidate.templateName(), interfaceCandidate.protocol())));
+					templatesCreated.add(interfaceCandidate.templateName());
+
+					if (sysInfo.getServiceDiscoveryInterfacePolicy() == ServiceDiscoveryInterfacePolicy.EXTENDABLE) {
+						final List<ServiceInterfaceTemplateProperty> templateProps = new ArrayList<>();
+						interfaceCandidate.properties().keySet().forEach(propName -> {
```
rbocsi, line 264: somewhere here you should check that at least one property is in the map

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -1,18 +1,443 @@
...
+					matching = false;
+				}
+
+				// Match against to version requirements
+				if (matching && !Utilities.isEmpty(filters.getVersions()) && !filters.getVersions().contains(serviceCandidate.getVersion())) {
+					matching = false;
+				}
+
+				// Match against to alive requirements
+				if (matching && filters.getAlivesAt() != null && serviceCandidate.getExpiresAt() != null && filters.getAlivesAt().isBefore(serviceCandidate.getExpiresAt())) {
```
rbocsi: i think the last condition is supposed to be the opposite.  There is no match if the expire time is before the alive time.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -1,18 +1,443 @@
...
+					}
+				}
+
+				if (matching) {
+					matchingIds.add(serviceCandidate.getId());
+				}
+			}
+
+			serviceInstanceRepo.deleteAllById(deleteIds);
+			final Page<ServiceInstance> page = serviceInstanceRepo.findAllByIdIn(matchingIds, pagination);
```
rbocsi, line 438: Not sure if needed but I put a flush after the delete just to be safe.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInterfaceTemplateDbService.java
```diff
@@ -1,18 +1,56 @@
...
 
 	//=================================================================================================
 	// methods
 
+	//-------------------------------------------------------------------------------------------------
+	public Optional<ServiceInterfaceTemplate> getByName(final String name) {
+		logger.debug("getByName started");
+		Assert.isTrue(!Utilities.isEmpty(name), "name is empty");
+
+		return templateRepo.findByName(name);
```
rbocsi: surround with the usual try-catch

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInterfaceTemplateDbService.java
```diff
@@ -1,18 +1,56 @@
...
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public List<ServiceInterfaceTemplateProperty> getPropertiesByName(final String name) {
+		logger.debug("getByName started");
```
rbocsi: name - log message mismatch

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInterfaceTemplateDbService.java
```diff
@@ -1,18 +1,56 @@
...
+		Assert.isTrue(!Utilities.isEmpty(name), "name is empty");
+
+		final Optional<ServiceInterfaceTemplate> templateOpt = templateRepo.findByName(name);
+
+		if (templateOpt.isPresent()) {
+			return templatePropsRepo.findByServiceInterfaceTemplate(templateOpt.get());
+		}
```
rbocsi: surround by the usual try-catch

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInterfaceTemplateDbService.java
```diff
@@ -1,18 +1,56 @@
...
+	//-------------------------------------------------------------------------------------------------
+	public Optional<ServiceInterfaceTemplate> getByName(final String name) {
+		logger.debug("getByName started");
+		Assert.isTrue(!Utilities.isEmpty(name), "name is empty");
+
+		return templateRepo.findByName(name);
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public List<ServiceInterfaceTemplateProperty> getPropertiesByName(final String name) {
```
rbocsi: I'd rename this method to getPropertiesByTemplateName, because at first I thought the name parameter is the name of a property, not a template.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ServiceDiscoveryService.java
```diff
@@ -0,0 +1,153 @@
...
+	public ServiceInstanceResponseDTO registerService(final ServiceInstanceRequestDTO dto, final String origin) {
+		logger.debug("registerService started");
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final ServiceInstanceRequestDTO normalized = validator.validateAndNormalizeRegisterService(dto, origin);
+		final String instanceId = ServiceInstanceIdUtils.calculateInstanceId(normalized.systemName(), normalized.serviceDefinitionName(), normalized.version());
+
+		try {
+			instanceDbService.deleteByInstanceId(instanceId);
+			final Entry<ServiceInstance, List<ServiceInstanceInterface>> instanceEntry = instanceDbService.createBulk(List.of(normalized)).getFirst();
```
rbocsi: This two lines should be in the db service in one transaction. Because if the new registration is failed, the old one does not come back.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ServiceDiscoveryService.java
```diff
@@ -0,0 +1,153 @@
...
+
+	//-------------------------------------------------------------------------------------------------
+	public ServiceInstanceListResponseDTO lookupServices(final String requesterSystemName, final ServiceInstanceLookupRequestDTO dto, final boolean verbose, final boolean restricted, final String origin) {
+		logger.debug("lookupServices started");
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final ServiceInstanceLookupRequestDTO normalized = validator.validateAndNormalizeLookupService(dto, origin);
+
+		if (restricted) {
+			for (MetadataRequirementDTO metadataReq : normalized.metadataRequirementsList()) {
```
rbocsi: Heresy :)
Missing final

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ServiceDiscoveryService.java
```diff
@@ -0,0 +1,153 @@
...
+				metadataReq.put(ServiceRegistryConstants.METADATA_KEY_UNRESTRICTED_DISCOVERY, true);
+			}
+		}
```
rbocsi, line 102: Missing case: if there is no metadataRequirements you have to create one with the restricted discovery requirement

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ServiceDiscoveryService.java
```diff
@@ -0,0 +1,153 @@
...
+						List.copyOf(servicesWithInterfaces.stream().map(e -> e.getKey().getSystem().getName()).collect(Collectors.toSet())),
+						null, null, null, null, null);
+
+				return dtoConverter.convertServiceInstanceListToDTO(servicesWithInterfaces, systemsWithDevices);
+			}
+
+			return dtoConverter.convertServiceInstanceListToDTO(servicesWithInterfaces, null);
+
+		} catch (final ForbiddenException ex) {
+			throw ex;
```
rbocsi: maybe add origin here too

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ServiceDiscoveryService.java
```diff
@@ -0,0 +1,153 @@
...
+		} catch (final InvalidParameterException ex) {
+			throw new InvalidParameterException(ex.getMessage(), origin);
+
+		} catch (final InternalServerError ex) {
+			throw new InternalServerError(ex.getMessage(), origin);
+		}
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public ServiceInstanceListResponseDTO lookupServices(final String requesterSystemName, final ServiceInstanceLookupRequestDTO dto, final boolean verbose, final boolean restricted, final String origin) {
```
rbocsi: why do you need the first parameter here?

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/dto/DTOConverter.java
```diff
@@ -169,6 +174,61 @@ public SystemListResponseDTO convertSystemListResponseDtoToTerse(final SystemLis
 
 	}
 
+	//-------------------------------------------------------------------------------------------------
+	public ServiceInstanceResponseDTO convertServiceInstanceEntityToDTO(final Entry<ServiceInstance, List<ServiceInstanceInterface>> instanceEntry, final Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>> systemTripet) {
```
rbocsi: typo: systemTripet => systemTriplet

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/model/ServiceLookupFilterModel.java
```diff
@@ -0,0 +1,155 @@
...
+	private Set<ServiceInterfacePolicy> policies = new HashSet<>();
+
+	//=================================================================================================
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	public ServiceLookupFilterModel(final ServiceInstanceLookupRequestDTO dto) {
+		Assert.notNull(dto, "ServiceInstanceLookupRequestDTO is null");
+
+		instanceIds = Utilities.isEmpty(dto.instanceIds()) ? new HashSet<>() : dto.instanceIds().stream().map(id -> id).collect(Collectors.toSet());
```
rbocsi: why do you need a map method call that doing nothing?
If no reason this line can replace with:
```
if (!Utilities.isEmpty(dto.instanceIds())) {
    instanceIds.addAll(dto.instanceIds());
}
```
The same is true to providerNames, serviceDefinitionNames and versions. And metadataRequirementsList. And interfaceTemplateNames and interfacePropertyRequirementsList

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/model/ServiceLookupFilterModel.java
```diff
@@ -0,0 +1,155 @@
...
+		providerNames = Utilities.isEmpty(dto.providerNames()) ? new HashSet<>() : dto.providerNames().stream().map(pn -> pn).collect(Collectors.toSet());
+		serviceDefinitionNames = Utilities.isEmpty(dto.serviceDefinitionNames()) ? new HashSet<>() : dto.serviceDefinitionNames().stream().map(sd -> sd).collect(Collectors.toSet());
+		versions = Utilities.isEmpty(dto.versions()) ? new HashSet<>() : dto.versions().stream().map(v -> v).collect(Collectors.toSet());
+		alivesAt = Utilities.isEmpty(dto.alivesAt()) ? null : Utilities.parseUTCStringToZonedDateTime(dto.alivesAt());
```
rbocsi: parse handles null or empty values too

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/model/ServiceLookupFilterModel.java
```diff
@@ -0,0 +1,155 @@
...
+	//=================================================================================================
+	// boilerplate
+
+	//-------------------------------------------------------------------------------------------------
+	public Set<String> getInstanceIds() {
+		return instanceIds;
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public void setInstanceIds(final Set<String> instanceIds) {
```
rbocsi: All the setters seems unnecessary to me. If you follow my advice regarding the constructor, all members (except policies) can be final. Policies can be final too if you use `addAll(<result of the ternary operator>)` on policies instead of assignment.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ServiceDiscoveryNormalization.java
```diff
@@ -0,0 +1,78 @@
...
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	public ServiceInstanceRequestDTO normalizeServiceInstanceRequestDTO(final ServiceInstanceRequestDTO dto) {
+		logger.debug("normalizeServiceInstanceRequestDTO started");
+		Assert.notNull(dto, "ServiceInstanceRequestDTO is null");
+
+		return new ServiceInstanceRequestDTO(
+				dto.systemName().trim(),
+				dto.serviceDefinitionName().trim().toLowerCase(),
```
rbocsi, line 38: We don't do this in case of system and device names, I think service definition names can be case sensitive too

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ServiceDiscoveryNormalization.java
```diff
@@ -0,0 +1,78 @@
...
+				versionNormalizer.normalize(dto.version()),
+				Utilities.isEmpty(dto.expiresAt()) ? "" : dto.expiresAt().trim(),
+				dto.metadata(),
+				Utilities.isEmpty(dto.interfaces()) ? new ArrayList<>()
+						: dto.interfaces()
+								.stream()
+								.map(i -> new ServiceInstanceInterfaceRequestDTO(
+										i.templateName().trim().toUpperCase(),
```
rbocsi: I'm not sure we want template names to be case insensitive.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ServiceDiscoveryNormalization.java
```diff
@@ -0,0 +1,78 @@
...
+		Assert.notNull(dto, "ServiceInstanceLookupRequestDTO is null");
+
+		return new ServiceInstanceLookupRequestDTO(
+				Utilities.isEmpty(dto.instanceIds()) ? new ArrayList<>() : dto.instanceIds().stream().map(id -> id.trim()).toList(),
+				Utilities.isEmpty(dto.providerNames()) ? new ArrayList<>() : dto.providerNames().stream().map(n -> n.trim()).toList(),
+				Utilities.isEmpty(dto.serviceDefinitionNames()) ? new ArrayList<>() : dto.serviceDefinitionNames().stream().map(sd -> sd.trim()).toList(),
+				Utilities.isEmpty(dto.versions()) ? new ArrayList<>() : dto.versions().stream().map(v -> versionNormalizer.normalize(v)).toList(),
+				Utilities.isEmpty(dto.alivesAt()) ? "" : dto.alivesAt().trim(),
+				Utilities.isEmpty(dto.metadataRequirementsList()) ? new ArrayList<>() : dto.metadataRequirementsList(),
+				Utilities.isEmpty(dto.interfaceTemplateNames()) ? new ArrayList<>() : dto.interfaceTemplateNames().stream().map(i -> i.trim().toUpperCase()).toList(),
```
rbocsi: see above about template name case sensitivity

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ServiceDiscoveryNormalization.java
```diff
@@ -0,0 +1,78 @@
...
+				Utilities.isEmpty(dto.interfacePropertyRequirementsList()) ? new ArrayList<>() : dto.interfacePropertyRequirementsList(),
+				Utilities.isEmpty(dto.policies()) ? new ArrayList<>() : dto.policies().stream().map(p -> p.trim().toUpperCase()).toList()
+		);
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public String normalizeServiceInstanceId(final String instanceId) {
+		logger.debug("normalizeServiceInstanceRequestDTO started");
```
rbocsi: wrong name in log message

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ServiceDiscoveryNormalization.java
```diff
@@ -0,0 +1,78 @@
...
+		Assert.isTrue(!Utilities.isEmpty(instanceId), "Service instance id is empty");
+
+		return instanceId.trim().toLowerCase();
```
rbocsi, line 76: see above about case sensitivity (system and service definition names)

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/utils/ServiceInstanceIdUtils.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.util.Assert;
+
+import eu.arrowhead.common.Utilities;
+
+public final class ServiceInstanceIdUtils {
+
+	//=================================================================================================
+	// members
+
+	private static final String delimiter = "-";
```
rbocsi: Let's choose a different delimiter, since system names and service definition names can contains dash.
Example:
system: a-b, service: c => a-b-c-1.0.0
system: a, service: b-c => a-b-c-1.0.0

Good candidates: /  // : ::

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/utils/ServiceInstanceIdUtils.java
```diff
@@ -0,0 +1,40 @@
...
+	public static String calculateInstanceId(final String systemName, final String serviceDefinitionName, final String version) {
+		Assert.isTrue(!Utilities.isEmpty(systemName), "systemName is empty");
+		Assert.isTrue(!Utilities.isEmpty(serviceDefinitionName), "serviceDefinitionName is empty");
+		Assert.isTrue(!Utilities.isEmpty(version), "version is empty");
+
+		return systemName + delimiter + serviceDefinitionName + delimiter + version;
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public static String retriveSystemNameFromInstaceId(final String instanceId) {
```
rbocsi: typo: retrive => retrieve

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/utils/ServiceInstanceIdUtils.java
```diff
@@ -0,0 +1,40 @@
...
+		Assert.isTrue(!Utilities.isEmpty(instanceId), "instanceId is empty");
+		return instanceId.split(delimiter)[0];
```
rbocsi: We should add another assert that a split to delimiter returns an array with length 3.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ServiceDiscoveryValidation.java
```diff
@@ -1,10 +1,238 @@
...
+import eu.arrowhead.serviceregistry.service.validation.version.VersionValidator;
+
 @Service
 public class ServiceDiscoveryValidation {
 
-	// TODO: every operation has a validation method that takes the input (DTO and/or params) and throws ArrowheadExceptions if anything is not valid.
-	// must be called from the service layer (example: ServiceDiscoveryService component)
+	//=================================================================================================
+	// members
+
```
rbocsi, line 34: We have to integrate the NameValidator that @Katinka666 implemented in the service-definition branch. We should use it on service definition names and i think template names as well.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ServiceDiscoveryValidation.java
```diff
@@ -1,10 +1,238 @@
...
+		if (Utilities.isEmpty(dto.serviceDefinitionName())) {
+			throw new InvalidParameterException("Service definition name is empty", origin);
+		}
+
+		if (dto.serviceDefinitionName().length() > ServiceRegistryConstants.SERVICE_DEFINITION_NAME_LENGTH) {
+			throw new InvalidParameterException("Service definition name is too long", origin);
+		}
+
+		// TODO validate service definition naming convention
```
rbocsi: ok, I see you want to do the same thing. :)

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ServiceDiscoveryValidation.java
```diff
@@ -1,10 +1,238 @@
...
+
+		if (!Utilities.isEmpty(dto.expiresAt())) {
+			try {
+				Utilities.parseUTCStringToZonedDateTime(dto.expiresAt());
+			} catch (final DateTimeException ex) {
+				throw new InvalidParameterException("Expiration time has an invalide time format", origin);
```
rbocsi: typo (or maybe french): invalide -> invalid
maybe also check that the expiration date is in the future

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ServiceDiscoveryValidation.java
```diff
@@ -1,10 +1,238 @@
...
+
+		if (!Utilities.isEmpty(dto.versions()) && Utilities.containsNullOrEmpty(dto.versions())) {
+			throw new InvalidParameterException("Version list contains null or empty element", origin);
+		}
+
+		if (!Utilities.isEmpty(dto.alivesAt())) {
+			try {
+				Utilities.parseUTCStringToZonedDateTime(dto.alivesAt());
+			} catch (final DateTimeException ex) {
+				throw new InvalidParameterException("Alive time has an invalide time format", origin);
```
rbocsi: french again :)

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ServiceDiscoveryValidation.java
```diff
@@ -1,10 +1,238 @@
...
+		validateRegisterService(dto, origin);
+
+		final ServiceInstanceRequestDTO normalized = normalizer.normalizeServiceInstanceRequestDTO(dto);
+
+		try {
+			versionValidator.validateNormalizedVersion(normalized.version());
+
+			normalized.interfaces().forEach(interfaceInstance -> {
+				final Optional<ServiceInterfaceTemplate> templateOpt = interfaceTemplateDbService.getByName(interfaceInstance.templateName());
+				if (templateOpt.isPresent() && !Utilities.isEmpty(interfaceInstance.protocol())) {
```
rbocsi: I think the second part of this condition is not needed here but in the first nested if. If we find the template we want to check the properties regardless of the protocol field.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ServiceDiscoveryValidation.java
```diff
@@ -1,10 +1,238 @@
...
+					if (!interfaceInstance.protocol().equals(templateOpt.get().getProtocol())) {
```
rbocsi: The second part of the previous if goes here as first part.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ServiceDiscoveryValidation.java
```diff
@@ -1,10 +1,238 @@
...
+					interfaceTemplateDbService.getPropertiesByName(interfaceInstance.templateName())
+							.forEach(templateProp -> {
+								final Object instanceProp = interfaceInstance.properties().get(templateProp.getPropertyName());
+
+								if (instanceProp == null && templateProp.isMandatory()) {
+									throw new InvalidParameterException(templateProp.getPropertyName() + " interface property is missing for " + templateProp.getServiceInterfaceTemplate().getName());
+								}
+
+								if (!Utilities.isEmpty(templateProp.getValidator())) {
+									final List<String> validatorWithArgs = Arrays.asList(templateProp.getValidator().split("\\|"));
```
rbocsi: Delimiter should be a constant.
I don't see why do you work with list when eventually you'll need an array.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ServiceDiscoveryValidation.java
```diff
@@ -1,10 +1,238 @@
...
+									final IPropertyValidator validator = interfacePropertyValidator.getValidator(PropertyValidatorType.valueOf(validatorWithArgs.get(0)));
+									if (validator != null) {
+										final Object normalizedProp = validator.validateNormalize(
+												instanceProp,
+												validatorWithArgs.size() <= 1 ? new String[0] : validatorWithArgs.subList(1, validatorWithArgs.size() - 1).toArray(new String[0]));
```
rbocsi: You can use Arrays.copyOfRange() instead of subList if you stick with the array.

rbocsi (changes requested):
see inline comments.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ServiceDiscoveryNormalization.java
rbocsi: should be lowercase

rbocsi: + task: database script currently uses uppercase template names as well. Please fix it.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/ServiceDiscoveryAPI.java
```diff
@@ -1,27 +1,144 @@
...
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_INTERNAL_SERVER_ERROR, description = Constants.SWAGGER_HTTP_500_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) })
+	})
+	@PostMapping(path = ServiceRegistryConstants.HTTP_API_OP_LOOKUP_PATH, consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
+	public @ResponseBody ServiceInstanceListResponseDTO lookup(
+									final HttpServletRequest httpServletRequest,
+									@RequestBody final  ServiceInstanceLookupRequestDTO dto,
+									@Parameter(
+											name =  "verbose",
+											description  = "Set true if you want the response to contains the system and device details. (It should be configured in the Application properties as well.)",
```
Katinka666: typo: to contain

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/utils/ServiceLookupPreprocessor.java
```diff
@@ -0,0 +1,32 @@
...
+		}
+
+		final Object boolObject = request.getAttribute(ServiceRegistryConstants.HTTP_ATTR_RESTRICTED_SERVICE_LOOKUP);
+		return boolObject != null && Boolean.valueOf(boolObject.toString());
```
Katinka666, line 30: Based on Rajmund's comment, wouldn't true be the default if the boolObject is null?

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ServiceDiscoveryService.java
```diff
@@ -0,0 +1,153 @@
...
+		logger.debug("lookupServices started");
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final ServiceInstanceLookupRequestDTO normalized = validator.validateAndNormalizeLookupService(dto, origin);
+
+		if (restricted) {
+			for (MetadataRequirementDTO metadataReq : normalized.metadataRequirementsList()) {
+				metadataReq.put(ServiceRegistryConstants.METADATA_KEY_UNRESTRICTED_DISCOVERY, true);
```
Katinka666, line 100: Souldn't the value of METADATA_KEY_UNRESTRICTED_DISCOVERY be "false" here, since the restricted flag was true?

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ServiceDiscoveryService.java
```diff
@@ -0,0 +1,153 @@
...
+		try {
+			final Page<Entry<ServiceInstance, List<ServiceInstanceInterface>>> servicesWithInterfaces = instanceDbService.getPageByFilters(
+					PageRequest.of(0, Integer.MAX_VALUE, Direction.DESC, ServiceInstance.DEFAULT_SORT_FIELD),
+					new ServiceLookupFilterModel(normalized));
+
+			if (verbose && !sysInfo.isDiscoveryVerbose()) {
+				throw new ForbiddenException("Verbose is not allowed", origin);
+			}
+
+			if (verbose) {
```
Katinka666, line 109: It can be combined with the previous expression:
```
if (verbose) {
   if (!sysInfo.isDiscoveryVerbose()) {
        throw new ForbiddenException("Verbose is not allowed", origin);
    }
//...
}
```

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/utils/ServiceLookupPreprocessor.java
rbocsi: No, if the boolObject is null it means the filter does not filled it because the policy is open.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ServiceDiscoveryService.java
rbocsi: The key means that the service can lookup without any restriction (even if the lookup is restricted).

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ServiceDiscoveryNormalization.java
borditamas: It has been decided afterward: All names should be lower-case #27

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInterfaceTemplateDbService.java
```diff
@@ -37,20 +38,34 @@ public Optional<ServiceInterfaceTemplate> getByName(final String name) {
...
+			logger.debug(ex);
+			throw new InternalServerError("Database operation error");
+		}
 	}
 
 	//-------------------------------------------------------------------------------------------------
-	public List<ServiceInterfaceTemplateProperty> getPropertiesByName(final String name) {
-		logger.debug("getByName started");
+	public List<ServiceInterfaceTemplateProperty> getPropertiesByTemplateName(final String name) {
+		logger.debug("getPropertiesByName started");
```
rbocsi: name still not matching

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/utils/ServiceInstanceIdUtils.java
```diff
@@ -24,9 +25,12 @@ public static String calculateInstanceId(final String systemName, final String s
 	}
 
 	//-------------------------------------------------------------------------------------------------
-	public static String retriveSystemNameFromInstaceId(final String instanceId) {
+	public static String retrieveSystemNameFromInstaceId(final String instanceId) {
 		Assert.isTrue(!Utilities.isEmpty(instanceId), "instanceId is empty");
-		return instanceId.split(delimiter)[0];
+
+		final String[] split = instanceId.split(delimiter);
+		Assert.isTrue(split.length == parts, "Invalid instaceId");
```
rbocsi: typo: instace => instance

rbocsi (changes requested):
I found some minor issues. You can merge after addressed them.

PR #34, opened 2024-10-17
Kj/service instances

Katinka666:


serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/ManagementAPI.java
```diff
@@ -376,15 +380,98 @@ public void removeServiceDefinitions(final @RequestParam List<String> names) {
...
+		final String origin = HttpMethod.POST.name() + " " + ServiceRegistryConstants.HTTP_API_MANAGEMENT_PATH + ServiceRegistryConstants.HTTP_API_OP_SERVICE_INSTANCE_PATH;
+		return mgmtService.createServiceInstances(dto, origin);
+	}
 
-	// update-service-instances PUT /service-instances(bulk) -> only metadata, expiresAt and interface can be changed
+	//-------------------------------------------------------------------------------------------------
+	@Operation(summary = "Returns the updated system entries")
+	@ApiResponses(value = {
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_OK, description = Constants.SWAGGER_HTTP_200_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = SystemListResponseDTO.class)) }),
```
rbocsi: return type is wrong here

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -154,6 +155,50 @@ public List<Entry<ServiceInstance, List<ServiceInstanceInterface>>> createBulk(f
 		}
 	}
 
+	//-------------------------------------------------------------------------------------------------
+	@Transactional(rollbackFor = ArrowheadException.class)
+	public List<Entry<ServiceInstance, List<ServiceInstanceInterface>>> updateBulk(final List<ServiceInstanceUpdateRequestDTO> dtos) {
+		logger.debug("createBulk started");
```
rbocsi: not matching with method name

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -154,6 +155,50 @@ public List<Entry<ServiceInstance, List<ServiceInstanceInterface>>> createBulk(f
...
+					instance.setMetadata(Utilities.toJson(dto.metadata()));
+					// set expires At
+					instance.setExpiresAt(Utilities.parseUTCStringToZonedDateTime(dto.expiresAt()));
+					// delete old interfaces
+					serviceInstanceInterfaceRepo.deleteAllByServiceInstance(instance);
+					// create new interfaces
+					final List<ServiceInstanceInterface> newInterfaces = createAndSaveInterfaces(dto.interfaces(), instance);
+
+					// save the new instance
+					serviceInstanceRepo.saveAndFlush(instance);
```
rbocsi: assign to instance to make sure instance's fields are up-to-date.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -301,6 +362,62 @@ private List<ServiceInstanceInterface> createInstanceInterfaces(final List<Servi
...
+				if (sysInfo.getServiceDiscoveryInterfacePolicy() == ServiceDiscoveryInterfacePolicy.RESTRICTED) {
+					// we don't create new template in case of restricted interface policy
+					throw new InvalidParameterException("Interface template not exists: " + dto.templateName());
+				} else {
+					// create new template
+					template = new ServiceInterfaceTemplate(dto.templateName(), dto.protocol());
+					// we add all the provided properties as mandatory properties in case of extendable interface policy
+					if (sysInfo.getServiceDiscoveryInterfacePolicy() == ServiceDiscoveryInterfacePolicy.EXTENDABLE) {
+						final List<ServiceInterfaceTemplateProperty> templateProps = new ArrayList<>();
+						for (Entry<String, Object> property : (List<Entry<String, Object>>) dto.properties()) {
```
rbocsi: missing final

rbocsi: dto.properties() is a `Map<String, Object>`. I'm not sure that the cast is valid. You can use entrySet() method on the map, or you can see what Tomi did in similar situation (foreach()).

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -301,6 +362,62 @@ private List<ServiceInstanceInterface> createInstanceInterfaces(final List<Servi
...
+					serviceInterfaceTemplateRepo.save(template);
+				}
+			} else {
+			// existing template
+				template = optionalTemplate.get();
+				// modify template protocol if necessary
+				if (template.getProtocol() != dto.protocol()) {
+					template.setProtocol(dto.protocol());
+					serviceInterfaceTemplateRepo.save(template);
+				}
```
rbocsi: you can't change an existing interface, because that can be used by other services as well. If the protocol in the dto is set (it is optional) and it does not match with the existing template's protocol, just throw an exception.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -301,6 +362,62 @@ private List<ServiceInstanceInterface> createInstanceInterfaces(final List<Servi
...
+				// we have to check, if all the mandatory properties are provided
+				final List<ServiceInterfaceTemplateProperty> mandatoryProperties = serviceInterfaceTemplatePropsRepo.findAllByServiceInterfaceTemplate(template)
+						.stream().filter(p -> p.isMandatory()).toList();
+				mandatoryProperties.forEach(p -> {
+					if (!dto.properties().containsKey(p.getPropertyName())) {
+						throw new InvalidParameterException("Mandatory interface property is missing: " + p.getPropertyName());
+					}
+				});
+			}
+			Assert.isTrue(Utilities.isEnumValue(dto.policy(), ServiceInterfacePolicy.class), "Invalid service interface policy");
```
rbocsi: This assert should be checked as soon as possible.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -301,6 +362,62 @@ private List<ServiceInstanceInterface> createInstanceInterfaces(final List<Servi
...
+				}
+
+				// we have to check, if all the mandatory properties are provided
+				final List<ServiceInterfaceTemplateProperty> mandatoryProperties = serviceInterfaceTemplatePropsRepo.findAllByServiceInterfaceTemplate(template)
+						.stream().filter(p -> p.isMandatory()).toList();
+				mandatoryProperties.forEach(p -> {
+					if (!dto.properties().containsKey(p.getPropertyName())) {
+						throw new InvalidParameterException("Mandatory interface property is missing: " + p.getPropertyName());
+					}
+				});
```
rbocsi: This can be one lambda expression. You can use foreach on the result of filter directly.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -241,6 +241,23 @@ public Optional<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAdd
...
 	}
 
+	//-------------------------------------------------------------------------------------------------
+	public List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> getByNameList(final List<String> names) {
+		logger.debug("getByName started");
+		Assert.isTrue(!Utilities.containsNullOrEmpty(names), "system name list contains null or empty");
+
+		List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> result = new ArrayList<>();
+
+		for (String name : names) {
```
rbocsi: missing final

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -241,6 +241,23 @@ public Optional<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAdd
 
 	}
 
+	//-------------------------------------------------------------------------------------------------
+	public List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> getByNameList(final List<String> names) {
+		logger.debug("getByName started");
+		Assert.isTrue(!Utilities.containsNullOrEmpty(names), "system name list contains null or empty");
+
+		List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> result = new ArrayList<>();
```
rbocsi: missing final

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/SystemDbService.java
```diff
@@ -241,6 +241,23 @@ public Optional<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAdd
...
+
+		for (String name : names) {
+			Optional<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> optionalTriple = getByName(name);
```
rbocsi: missing final

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -289,6 +304,110 @@ public void removeSystems(final List<String> names, final String origin) {
...
+
+	//-------------------------------------------------------------------------------------------------
+	public ServiceInstanceListResponseDTO createServiceInstances(final ServiceInstanceCreateListRequestDTO dto, final String origin) {
+		logger.debug("createServiceInstances started");
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final List<ServiceInstanceRequestDTO> normalized = validator.validateAndNormalizeCreateServiceInstances(dto, origin);
+
+		final List<Entry<ServiceInstance, List<ServiceInstanceInterface>>> instanceEntries;
+		final List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> systemTriplets;
```
rbocsi, line 317: These can't be final because don't initialized and get values later.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -289,6 +304,110 @@ public void removeSystems(final List<String> names, final String origin) {
...
+
+		try {
+			synchronized (LOCK) {
```
rbocsi: I think this syncronized here is unnecessary.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -289,6 +304,110 @@ public void removeSystems(final List<String> names, final String origin) {
...
+	public ServiceInstanceListResponseDTO updateServiceInstance(final ServiceInstanceUpdateListRequestDTO dto, final String origin) {
+		logger.debug("updateServiceInstance started");
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final List<ServiceInstanceUpdateRequestDTO> normalized = validator.validateAndNormalizeUpdateServiceInstances(dto, origin);
+
+		try {
+
+			final List<Entry<ServiceInstance, List<ServiceInstanceInterface>>> updatedEntries;
+			final List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> systemTriplets;
```
rbocsi, line 345: unnecessary finals

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -289,6 +304,110 @@ public void removeSystems(final List<String> names, final String origin) {
...
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final ServiceInstanceQueryRequestDTO normalized = validator.validateAndNormalizeQueryServiceInstances(dto, origin);
+
+		try {
+			final PageRequest pageRequest = pageService.getPageRequest(normalized.pagination(), Direction.DESC, ServiceInstance.SORTABLE_FIELDS_BY, ServiceInstance.DEFAULT_SORT_FIELD, origin);
+
+			Page<Entry<ServiceInstance, List<ServiceInstanceInterface>>> servicesWithInterfaces;
+			Page<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> systemsWithDevices = null;
+			synchronized (LOCK) {
```
rbocsi: I think this syncronized here is unnecessary.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -289,6 +304,110 @@ public void removeSystems(final List<String> names, final String origin) {
...
+		logger.debug("updateServiceInstance started");
+		Assert.isTrue(!Utilities.isEmpty(origin), "origin is empty");
+
+		final List<ServiceInstanceUpdateRequestDTO> normalized = validator.validateAndNormalizeUpdateServiceInstances(dto, origin);
+
+		try {
+
+			final List<Entry<ServiceInstance, List<ServiceInstanceInterface>>> updatedEntries;
+			final List<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> systemTriplets;
+			synchronized (LOCK) {
```
rbocsi: I think this syncronized here is unnecessary.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/api/http/ManagementAPI.java
```diff
@@ -376,15 +380,98 @@ public void removeServiceDefinitions(final @RequestParam List<String> names) {
...
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) }),
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_INTERNAL_SERVER_ERROR, description = Constants.SWAGGER_HTTP_500_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) })
+	})
+	@PostMapping(path = ServiceRegistryConstants.HTTP_API_OP_SERVICE_INSTANCE_QUERY_PATH, consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
+	public @ResponseBody ServiceInstanceListResponseDTO queryServiceInstances(
+			@RequestBody final ServiceInstanceQueryRequestDTO dto,
+			@Parameter(
+					name =  "verbose",
+					description  = "Set true if you want the response to contain the system and device details. (It should be configured in the Application properties as well.)",
```
rbocsi: The part between the parantheses are not true in a management service operation.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
```diff
@@ -74,8 +84,13 @@ public class ManagementService {
 	@Autowired
 	private DTOConverter dtoConverter;
 
+	@Autowired
+	private ServiceInstanceDbService instanceDbService;
+
 	private final Logger logger = LogManager.getLogger(this.getClass());
 
+	private static final Object LOCK = new Object();
```
rbocsi: I think you don't need this lock.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/dto/DTOConverter.java
```diff
@@ -224,6 +227,29 @@ public ServiceInstanceResponseDTO convertServiceInstanceEntityToDTO(final Entry<
 		);
 	}
 
+	//-------------------------------------------------------------------------------------------------
+	public ServiceInstanceListResponseDTO convertServiceInstancePageToDTO(
+																  final Page<Entry<ServiceInstance, List<ServiceInstanceInterface>>> servicesWithInterfaces,
+																  final Iterable<Triple<System, List<SystemAddress>, Entry<Device, List<DeviceAddress>>>> systemsWithDevices) {
+		logger.debug("convertServiceInstanceListToDTO started...");
+
+		final List<ServiceInstanceResponseDTO> entries = new ArrayList<>();
```
rbocsi: you can initialize the arraylist capacity here.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/dto/DTOConverter.java
```diff
@@ -288,6 +314,20 @@ public KeyValuesDTO convertConfigMapToDTO(final Map<String, String> map) {
 
 	}
 
+	//-------------------------------------------------------------------------------------------------
+	public ServiceLookupFilterModel convertServiceInstanceQueryRequestDtoToFilterModel(final ServiceInstanceQueryRequestDTO dto) {
+		return new ServiceLookupFilterModel(new ServiceInstanceLookupRequestDTO(
```
rbocsi, line 321: missing logger debug message

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ManagementNormalization.java
```diff
@@ -163,6 +168,51 @@ public List<String> normalizeRemoveServiceDefinitions(final List<String> names)
...
+	public List<ServiceInstanceUpdateRequestDTO> normalizeUpdateServiceInstances(final ServiceInstanceUpdateListRequestDTO dto) {
+		logger.debug("normalizeUpdateServiceInstances started");
+		Assert.notNull(dto, "ServiceInstanceUpdateListRequestDTO is null");
+
+		return dto.instances().stream().map(i -> normalizeServiceInstanceUpdateRequestDTO(i)).collect(Collectors.toList());
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public List<String> normalizeRemoveServiceInstances(final List<String> instanceIds) {
+		logger.debug("normalizeUpdateServiceInstanceIds started");
```
rbocsi: message not match with method name

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ManagementNormalization.java
```diff
@@ -200,4 +250,71 @@ public List<String> normalizeRemoveInterfaceTemplates(final List<String> names)
 				.map(n -> nameNormalizer.normalize(n))
 				.collect(Collectors.toList());
 	}
+
+	//=================================================================================================
+	// assistant methods
+
+	//-------------------------------------------------------------------------------------------------
+	public ServiceInstanceRequestDTO normalizeServiceInstanceRequestDTO(final ServiceInstanceRequestDTO dto) {
```
rbocsi: it this public? because it is in the assistant methods part.

also missing starter debug message

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ManagementNormalization.java
```diff
@@ -200,4 +250,71 @@ public List<String> normalizeRemoveInterfaceTemplates(final List<String> names)
...
+				// interfaces
+				dto.interfaces()
+					.stream()
+					.map(i -> interfaceNormalizer.normalizeInterfaceDTO(i))
+					.toList()
+		);
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public ServiceInstanceUpdateRequestDTO normalizeServiceInstanceUpdateRequestDTO(final ServiceInstanceUpdateRequestDTO dto) {
```
rbocsi: it this public? because it is in the assistant methods part.

also missing starter debug message

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/normalization/ManagementNormalization.java
```diff
@@ -200,4 +250,71 @@ public List<String> normalizeRemoveInterfaceTemplates(final List<String> names)
...
+
+		return new ServiceInstanceUpdateRequestDTO(
+				// instance id
+				dto.instanceId().trim(),
```
rbocsi: you should use a name normalizer on this (beacuse system name and service definition name are normalized, and this one contains them)

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
```diff
@@ -497,6 +505,281 @@ public List<String> validateAndNormalizeRemoveSystems(final List<String> origina
...
+
+		if (dto == null) {
+			throw new InvalidParameterException("Request payload is missing", origin);
+		}
+
+		if (Utilities.isEmpty(dto.instances())) {
+			throw new InvalidParameterException("Request payload is empty", origin);
+		}
+
+		for (ServiceInstanceRequestDTO instance : dto.instances()) {
```
rbocsi: missing final

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
```diff
@@ -497,6 +505,281 @@ public List<String> validateAndNormalizeRemoveSystems(final List<String> origina
...
+		if (dto == null) {
+			throw new InvalidParameterException("Request payload is missing", origin);
+		}
+
+		if (Utilities.isEmpty(dto.instances())) {
+			throw new InvalidParameterException("Request payload is empty", origin);
+		}
+
+		List<String> instanceIds = new ArrayList<>();
+		for (ServiceInstanceUpdateRequestDTO instance : dto.instances()) {
```
rbocsi: missing final

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
```diff
@@ -497,6 +505,281 @@ public List<String> validateAndNormalizeRemoveSystems(final List<String> origina
...
+
+		if (dto == null) {
+			throw new InvalidParameterException("Request payload is missing", origin);
+		}
+
+		if (Utilities.isEmpty(dto.instances())) {
+			throw new InvalidParameterException("Request payload is empty", origin);
+		}
+
+		List<String> instanceIds = new ArrayList<>();
```
rbocsi: should be set

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
```diff
@@ -497,6 +505,281 @@ public List<String> validateAndNormalizeRemoveSystems(final List<String> origina
...
+		for (ServiceInstanceUpdateRequestDTO instance : dto.instances()) {
+
+			// instance id
+			if (Utilities.isEmpty(instance.instanceId())) {
+				throw new InvalidParameterException("Instance id is empty");
+			}
+
+			if (instanceIds.contains(instance.instanceId())) {
```
rbocsi: normalize instanceid before checking for duplication (to avoid duplicates that only different because of lowercase-uppercase letters) and always add the normalized value to the instanceIds.

rbocsi (changes requested):
see inline comments

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/ManagementService.java
Katinka666: What happens if I create the instanceEntries (line 321), than someone deletes some, so the systemTriplets (line 322) will not contain the matching system? Sould I check this in the dtoConverter?

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
Katinka666: Sould I do this change in other methods too where we perform duplication check?

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
Katinka666: Next time I will not forget about them!

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/service/validation/ManagementValidation.java
rbocsi: If somewhere we check for duplicates before normalization, then yes. We introduce this lowercase transformation not too long ago, so it is possible we don't change duplication check code.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -396,22 +398,20 @@ private List<ServiceInstanceInterface> createAndSaveInterfaces(final List<Servic
 			} else {
 			// existing template
 				template = optionalTemplate.get();
-				// modify template protocol if necessary
-				if (template.getProtocol() != dto.protocol()) {
-					template.setProtocol(dto.protocol());
-					serviceInterfaceTemplateRepo.save(template);
+				if (!template.getProtocol().equals(dto.protocol())) {
```
rbocsi: - only have to check if dto.protocol() is not null or empty (when using existing templates, the protocol is not a mandatory field in the dto).
- use equalsIgnoreCase() instead of equals() just to be safe

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -396,22 +398,20 @@ private List<ServiceInstanceInterface> createAndSaveInterfaces(final List<Servic
...
+					throw new InvalidParameterException("The protocol can not be overwritten");
```
rbocsi: need a better exception message. We should use what Tomi used in a similar case:
`"Interface has different protocol than " + dto.templateName() + " template"`

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -396,22 +398,20 @@ private List<ServiceInstanceInterface> createAndSaveInterfaces(final List<Servic
...
-				final List<ServiceInterfaceTemplateProperty> mandatoryProperties = serviceInterfaceTemplatePropsRepo.findAllByServiceInterfaceTemplate(template)
-						.stream().filter(p -> p.isMandatory()).toList();
-				mandatoryProperties.forEach(p -> {
-					if (!dto.properties().containsKey(p.getPropertyName())) {
-						throw new InvalidParameterException("Mandatory interface property is missing: " + p.getPropertyName());
-					}
+				serviceInterfaceTemplatePropsRepo.findAllByServiceInterfaceTemplate(template)
+						.stream()
+						.filter(p -> p.isMandatory())
+						.toList()
```
rbocsi: don't need this line. You can use forEach on a stream.

rbocsi (changes requested):
see new comments

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -398,14 +398,13 @@ private List<ServiceInstanceInterface> createAndSaveInterfaces(final List<Servic
 			} else {
 			// existing template
 				template = optionalTemplate.get();
-				if (!template.getProtocol().equals(dto.protocol())) {
-					throw new InvalidParameterException("The protocol can not be overwritten");
+				if (!Utilities.isEmpty(dto.protocol())) {
```
rbocsi: This is not what I meant. In this case the protocol is not mandatory, but it can be specified. If it specified it has to be the same (using equalsIgnoreCase()) than the protocol in the existing template. So you have to combine the two condition.

serviceregistry/src/main/java/eu/arrowhead/serviceregistry/jpa/service/ServiceInstanceDbService.java
```diff
@@ -398,14 +398,13 @@ private List<ServiceInstanceInterface> createAndSaveInterfaces(final List<Servic
...
+					throw new InvalidParameterException("The protocol for template with name " + dto.templateName()  + " already exists: " + template.getProtocol());
```
rbocsi: use the error message that Tomi used.

rbocsi (changes requested):
see comments

rbocsi (approved):

borditamas (approved):

PR #56, opened 2025-02-11
Rb/authentication

rbocsi:
MQTT endpoints are still missing


authentication/src/main/java/eu/arrowhead/authentication/AuthenticationSystemInfo.java
```diff
@@ -0,0 +1,206 @@
...
+	// assistant methods
+
+	//-------------------------------------------------------------------------------------------------
+	@Override
+	protected void customInit() {
+		if (Utilities.isEmpty(authenticationSecretKey)) {
+			throw new InvalidParameterException("'authenticationSecretKey' is missing or empty");
+		}
+
+		if (this.getAuthenticationPolicy() != AuthenticationPolicy.INTERNAL) {
```
Katinka666, line 120: What is the purpose of this check? The getAuthenticationPolicy() method always returns INTERNAL.

authentication/src/main/java/eu/arrowhead/authentication/api/http/IdentityAPI.java
```diff
@@ -0,0 +1,135 @@
...
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_UNAUTHORIZED, description = Constants.SWAGGER_HTTP_401_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) }),
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_INTERNAL_SERVER_ERROR, description = Constants.SWAGGER_HTTP_500_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) }),
+			@ApiResponse(responseCode = Constants.HTTP_STATUS_SERVICE_UNAVAILABLE, description = Constants.SWAGGER_HTTP_503_MESSAGE, content = {
+					@Content(mediaType = MediaType.APPLICATION_JSON_VALUE, schema = @Schema(implementation = ErrorMessageDTO.class)) })
+	})
+	@PostMapping(path = AuthenticationConstants.HTTP_API_OP_CHANGE_PATH, consumes = MediaType.APPLICATION_JSON_VALUE)
+	public void change(@RequestBody final IdentityChangeRequestDTO dto) {
+		logger.debug("logout started...");
```
Katinka666: typo in comment: change started

authentication/src/main/java/eu/arrowhead/authentication/api/http/utils/SystemNamePreprocessor.java
```diff
@@ -0,0 +1,30 @@
...
+	//=================================================================================================
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	public String process(final HttpServletRequest request, final String origin) throws InvalidParameterException {
+		logger.debug("process started");
+
+		final String name = HttpUtilities.acquireName(request, origin);
+
+		return name;
```
Katinka666: can be shorter: return HttpUtilities.acquireName(request, origin);

authentication/src/main/java/eu/arrowhead/authentication/http/filter/InternalAuthenticationFilter.java
```diff
@@ -0,0 +1,108 @@
...
+				final Optional<ActiveSession> sessionOpt = dbService.getSessionByToken(token.trim());
+				if (sessionOpt.isEmpty()) {
+					throw new AuthException("Invalid identity token");
+				}
+
+				final ActiveSession session = sessionOpt.get();
+				final ZonedDateTime now = Utilities.utcNow();
+				if (session.getExpirationTime().isBefore(now)) {
+					// expired session
+					throw new AuthException("Invalid identity token");
```
Katinka666, line 65: I think it might be useful for the user, if this error message is rather "Expired identity token". But I am not sure, maybe we souldn't tell this information?

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
```diff
@@ -0,0 +1,595 @@
...
+			logger.error(ex.getMessage());
+			logger.debug(ex);
+			throw new InternalServerError("Database operation error");
+		}
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	@Transactional(rollbackFor = ArrowheadException.class)
+	public ActiveSession createOrUpdateSession(final System system, final String token) {
+		logger.debug("getPasswordAuthenticationBySystem started...");
```
Katinka666: Debug message does not match with method name

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
```diff
@@ -0,0 +1,595 @@
...
+		} catch (final Exception ex) {
+			logger.error(ex.getMessage());
+			logger.debug(ex);
+			throw new InternalServerError("Database operation error");
+		}
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public List<System> getSystemsByNames(final List<String> names, final boolean strict) {
+		logger.debug("getSystemsByNamesBeforeUpdate started...");
```
Katinka666: Debug message does not match with method name

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
```diff
@@ -0,0 +1,595 @@
...
+		List<IdentityData> identityList;
+		List<System> systems;
+		synchronized (SYSTEM_LOCK) {
+			try {
+				checkSystemNamesNotExist(dto.identities()); // checks if none of the system names exist (system name has to be unique)
+
+				// writing the system entities to the database
+				identityList = createSystemEntitiesAndIdentityList(requester, dto.authenticationMethod(), dto.identities());
+				systems = systemRepository.saveAllAndFlush(identityList.stream().map(id -> id.system()).toList());
+				identityList = updateSystemEntitiesInIdentityList(identityList, systems);
```
Katinka666, line 143: Why is it necessary? I belive that it does something to the entities, but I haven't figured out what.

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
```diff
@@ -0,0 +1,595 @@
...
+				.map(c -> new IdentityData(
+						new System(c.systemName(), method.type(), c.sysop(), requester),
+						c.credentials(),
+						c.sysop()))
+				.collect(Collectors.toList());
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	// expects the same order in the two list
+	private List<IdentityData> updateSystemEntitiesInIdentityList(final List<IdentityData> identityList, final List<System> systems) {
```
Katinka666, line 462: Since it expects the same order in the two lists, I think a Map<IdentityData, System> would be better parameter instead of the two lists, because it forces the pairing. But perhaps this is not so important for an assistant method.

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
```diff
@@ -0,0 +1,595 @@
...
+		for (int i = 0; i < identityList.size(); ++i) {
+			result.add(new IdentityData(systems.get(i), identityList.get(i).credentials(), identityList.get(i).sysop()));
+		}
+
+		return result;
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	// expects the same order in the two list
+	private List<System> updateSystemListWithExtras(final List<System> systems, final List<String> extras) {
```
Katinka666, line 476: Same as above: I think a Map<System, String> would be a better parameter.

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
```diff
@@ -0,0 +1,595 @@
...
+		logger.debug("updateSystemListWithExtras started");
+		Assert.isTrue(extras.size() == systems.size(), "The two list has different size");
+
+		final List<System> result = new ArrayList<>(systems.size());
```
Katinka666, line 480: Why not modify the original list instead of creating a new one?
```
for (int i = 0; i < systems.size(); ++i) {
	systems.get(i).setExtra(extras.get(i));
}
return systems;
```

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
```diff
@@ -0,0 +1,595 @@
...
+		synchronized (SESSION_LOCK) {
+			try {
+				final Optional<ActiveSession> sessionOpt = asRepository.findBySystem(system);
+				final ZonedDateTime now = Utilities.utcNow();
+				final ZonedDateTime expirationTime = identityTokenDuration > 0 ? now.plusSeconds(identityTokenDuration) : now.plusYears(AuthenticationConstants.INFINITE_TOKEN_DURATION);
+
+				final ActiveSession session = sessionOpt.isPresent() ? sessionOpt.get() : new ActiveSession(system, token, now, expirationTime);
+				if (sessionOpt.isPresent()) {
+					if (now.isAfter(session.getExpirationTime())) {
+						// session is already expired, just the record is not removed => re-using it
```
Katinka666: **but** the record is not removed

authentication/src/main/java/eu/arrowhead/authentication/method/password/PasswordAuthenticationMethodInputNormalizer.java
```diff
@@ -0,0 +1,34 @@
...
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	@Override
+	public Map<String, String> normalizeCredentials(final Map<String, String> credentials) throws InternalServerError {
+		logger.debug("PasswordAuthenticationMethodInputNormalizer.normalizeCredentials started...");
+		Assert.notNull(credentials, "credentials is null");
+		Assert.isTrue(!Utilities.isEmpty(credentials.get(PasswordAuthenticationMethod.KEY_PASSWORD)), "password field is missing or empty");
+
+		return Map.of(PasswordAuthenticationMethod.KEY_PASSWORD, credentials.get(PasswordAuthenticationMethod.KEY_PASSWORD).trim());
```
Katinka666, line 32: Won't it mislead the user if we trim the password?

authentication/src/main/java/eu/arrowhead/authentication/service/validation/IdentityValidation.java
```diff
@@ -0,0 +1,138 @@
...
+		} catch (final InternalServerError ex) {
+			throw new InternalServerError(ex.getMessage(), origin);
+		}
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public IdentityChangeRequestDTO validateAndNormalizeChangeServicePhase2(final IdentityChangeRequestDTO dto, final AuthenticationMethod authenticationMethod, final String origin) {
+		logger.debug("validateAndNormalizeChangeServicePhase2 started...");
+
+		// at this point only new credential are not validated or normalized
```
Katinka666: typo: credential**s**

serviceregistry/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -15,7 +15,7 @@
     "description": "A description for 'authentication.policy'"
   },
   {
-    "name": "authenticator.secret.key",
+    "name": "authenticator.secret.keys",
     "type": "java.lang.String",
     "description": "A description for 'authenticator.secret.key'"
```
Katinka666: Maybe change the name here too (to keys)

Katinka666 (commented):
Found some typos, the others are rather suggestions and questions

authentication/src/main/java/eu/arrowhead/authentication/AuthenticationSystemInfo.java
rbocsi: You are right. I've override the getAuthenticationPolicy(), because INTERNAL is the only valid value in this case and I thought that with this I don't need the authentication.policy property at all . But later I realized that I can't remove authentication.policy, because in some bean creation it is referenced directly (can't use methods in EL expressions) and has a default value (CERTIFICATE). To maintain consistency I've put this check, but I forgot to remove the override.

So now, I fixed this issue with removing the overriding method.

authentication/src/main/java/eu/arrowhead/authentication/api/http/IdentityAPI.java
rbocsi: fixed

authentication/src/main/java/eu/arrowhead/authentication/http/filter/InternalAuthenticationFilter.java
rbocsi: Intentional. In security case, we usually tell as little information as possible.

authentication/src/main/java/eu/arrowhead/authentication/api/http/utils/SystemNamePreprocessor.java
rbocsi: fixed

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
rbocsi: fixed

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
rbocsi: fixed

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
rbocsi: All save... methods in repositories returns the saved entities (with new id in case of new records, changed updatedAt field, etc.). Honestly, I don't know whether the input object is changed or not (could be implementation-specific). So to be safe, I just merge the actual systems back to the identityList (assuming that the order is not changed)

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
rbocsi: When I use this method I have two lists. One come from the db repository. I'd have to create the map there (assuming the same ordering). It is just more work with nothing to gain.

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
rbocsi: In that case, the implementors of IAuthenticationMethodDbService would have more work to provide the necessary map. I've put a comment to the interface method for the ordering condition and I think that is enough.

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
rbocsi: It is just a habit, I think.

authentication/src/main/java/eu/arrowhead/authentication/jpa/service/IdentityDbService.java
rbocsi: fixed

authentication/src/main/java/eu/arrowhead/authentication/method/password/PasswordAuthenticationMethodInputNormalizer.java
rbocsi: I don't want passwords that contains leading/trailing white spaces. Also, it won't mislead the user, because both login, create, update and change operations are using the same normalization.

authentication/src/main/java/eu/arrowhead/authentication/service/validation/IdentityValidation.java
rbocsi: fixed

serviceregistry/src/main/resources/META-INF/additional-spring-configuration-metadata.json
rbocsi: fixed

Katinka666 (approved):

serviceregistry/src/main/resources/application.properties
```diff
@@ -32,10 +32,17 @@ mqtt.client.password=123456
 domain.name=127.0.0.1
 
 # Method to authenticate service requester systems
-authentication.policy=declared
+authentication.policy=outsourced
```
borditamas: Default value in github should be `declared` I think

borditamas (approved):
Approved with one comment

serviceregistry/src/main/resources/application.properties
rbocsi: fixed

rbocsi:
This one should only merged after the get-config rework PR.

PR #67, opened 2025-04-08
Authorization skeleton

rbocsi:


borditamas (approved):

consumer-authorization/src/main/java/eu/arrowhead/authorization/jpa/entity/UsageLimitedToken.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public int getUsageLimit() {
+		return usageLimit;
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	public void setUsageLimit(final int usageLimit) {
+		this.usageLimit = usageLimit;
```
Katinka666, line 85: It the usage limit changes, the usage left might change as well.

Katinka666 (approved):
Approved, but please see the one comment I left.

consumer-authorization/src/main/java/eu/arrowhead/authorization/jpa/entity/UsageLimitedToken.java
rbocsi: These setters are for the Hibernate to create an entity from a database record. It would be confusing if they are doing something more than setting the appropriate field.

PR #72, opened 2025-04-30
Rb/authorization policy

rbocsi:


consumer-authorization/src/main/java/eu/arrowhead/authorization/jpa/service/AuthorizationPolicyDbService.java
```diff
@@ -0,0 +1,628 @@
...
+		try {
+			// check existing record
+			final Optional<AuthProviderPolicyHeader> existing = providerHeaderRepository.findByCloudAndProviderAndTargetTypeAndTarget(
+					request.cloud(),
+					request.provider(),
+					request.targetType(),
+					request.target());
+
+			if (existing.isPresent()) {
+				throw new InvalidParameterException("A provider level policy is already existing for this cloud / provider / target type / target combination: "
```
borditamas: Shouldn't this override the existing rules? The non-management endpoints usually works that way in order to make the application system's life easier

rbocsi: In SR, we don't override systems and devices. We check if the requester wants to register the same thing (with the same attributes) and if the answer is yes, we just returned the already existing entity. 

So for override, no. But I agree that this should work like that: compare the existing records with the specified one and if the two are identitcal we should return the old one (in this case the status code should be 200 instead of 201).

borditamas: agree

consumer-authorization/src/main/java/eu/arrowhead/authorization/service/AuthorizationPolicyEngine.java
```diff
@@ -0,0 +1,191 @@
...
+import eu.arrowhead.common.http.ArrowheadHttpService;
+import eu.arrowhead.common.service.validation.MetadataRequirementsMatcher;
+import eu.arrowhead.dto.AuthorizationPolicyDTO;
+import eu.arrowhead.dto.SystemListResponseDTO;
+import eu.arrowhead.dto.SystemLookupRequestDTO;
+import eu.arrowhead.dto.enums.AuthorizationLevel;
+import eu.arrowhead.dto.enums.AuthorizationPolicyType;
+
+@Service
+public class AuthorizationPolicyEngine {
```
borditamas: I prefer to have only the service layer classes in the `eu.arrowhead.authorization.service` package and place anything else into an `eu.arrowhead.authorization.service.utils`  or `eu.arrowhead.authorization.service.<specific-name>` package as we have done so far.

serviceorchestration-dynamic/src/main/java/eu/arrowhead/serviceorchestration/service/utils/LocalServiceOrchestration.java
```diff
@@ -585,6 +585,7 @@ private OrchestrationResponseDTO convertToOrchestrationResponse(final List<Orche
 		final List<OrchestrationResultDTO> results = candidates.stream()
 				.map(c -> new OrchestrationResultDTO(
 						c.getServiceInstance().instanceId(),
+						null, // local cloud
```
borditamas: `DTODefaults.DEFAULT_CLOUD` maybe?

borditamas (changes requested):
see inline comments

consumer-authorization/src/main/java/eu/arrowhead/authorization/api/http/AuthorizationAPI.java
```diff
@@ -0,0 +1,161 @@
...
+		logger.debug("lookup started");
+
+		final String origin = HttpMethod.POST.name() + " " + AuthorizationConstants.HTTP_API_AUTHORIZATION_PATH + AuthorizationConstants.HTTP_API_OP_LOOKUP_PATH;
+		final String provider = sysNamePreprocessor.process(httpServletRequest, origin);
+
+		return authService.lookupOperation(provider, dto, origin);
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	@Operation(summary = "Returns the whether a consumer has access to a specified service/service operation/event type.")
```
Katinka666: typo: "the" is unnecessary

consumer-authorization/src/main/java/eu/arrowhead/authorization/api/http/AuthorizationManagementAPI.java
```diff
@@ -0,0 +1,150 @@
...
+	public @ResponseBody AuthorizationPolicyListResponseDTO queryPolicies(@RequestBody final AuthorizationQueryRequestDTO dto) {
+		logger.debug("queryPolicies started");
+
+		final String origin = HttpMethod.POST.name() + " " + AuthorizationConstants.HTTP_API_MANAGEMENT_PATH + AuthorizationConstants.HTTP_API_OP_QUERY_PATH;
+
+		return mgmtService.queryPoliciesOperation(dto, origin);
+	}
+
+	//-------------------------------------------------------------------------------------------------
+	@Operation(summary = "Returns the whether consumers has access to specified services/service operations/event types.")
```
Katinka666: same typo here

Katinka666 (approved):
only found a minor typo

borditamas (approved):

PR #86, opened 2025-08-06
DynamicOrch unit tests

borditamas:


serviceorchestration-dynamic/src/main/java/eu/arrowhead/serviceorchestration/service/model/validation/OrchestrationFormNormalization.java
```diff
@@ -58,6 +58,10 @@ public void normalizeOrchestrationForm(final OrchestrationForm form) {
 			form.setOperations(form.getOperations().stream().map(op -> serviceOpNameNormalizer.normalize(op)).toList());
 		}
 
+		if (!Utilities.isEmpty(form.getVersions())) {
+			form.setVersions(form.getVersions().stream().map(v -> v.trim()).toList());
```
rbocsi: There is a version normalizer component in common. It does more than trim. For example, "1" -> "1.0.0"

serviceorchestration-dynamic/src/main/java/eu/arrowhead/serviceorchestration/service/thread/PushOrchestrationWorker.java
```diff
@@ -0,0 +1,206 @@
...
+	private UUID jobId;
+
+	private final Logger logger = LogManager.getLogger(this.getClass());
+
+	//=================================================================================================
+	// methods
+
+	//-------------------------------------------------------------------------------------------------
+	public PushOrchestrationWorker(final UUID jobid) {
+		jobId = jobid;
```
rbocsi: This is ugly and dangerous (and probably not intentional). Should be `this.jobId = jobId;` and of course change the input to jobId.

rbocsi (changes requested):
Only reviewed the normal code, not the test code. Although I've run the tests and they are ok.

However, I find some bugs (fixes may need to modify tests).

rbocsi (approved):

serviceorchestration-dynamic/src/main/java/eu/arrowhead/serviceorchestration/jpa/service/OrchestrationLockDbService.java
```diff
@@ -57,7 +57,7 @@ public List<OrchestrationLock> create(final List<OrchestrationLock> candidates)
 	public List<OrchestrationLock> getByServiceInstanceId(final List<String> ids) {
 		logger.debug("getByServiceInstanceId started...");
 		Assert.isTrue(!Utilities.isEmpty(ids), "Service instance id list is empty");
-		Assert.isTrue(!Utilities.containsNull(ids), "Service instance id list contains null element");
+		Assert.isTrue(!Utilities.containsNullOrEmpty(ids), "Service instance id list contains null element");
```
Katinka666: This is not that important but the messange should be `contains null or empty element` then

Katinka666 (approved):
The tests have run for me too, please check the only comment I left.

